PR #25, opened 2022-05-23
Implementation of Spline reading for DxfSectionReaderBase

michal-kindl:
I was inspired by MLine Entity implementation and tried to follow implementation rules as in other Entities.
I am not sure how to set KnotParametrization flag. I see intensive development here, so I hope it can help.

### DIFF
--- a/ACadSharp/IO/Templates/CadSplineTemplate.cs
+++ b/ACadSharp/IO/Templates/CadSplineTemplate.cs
@@ -1,0 +1,86 @@
+﻿using ACadSharp.Entities;
+using System;
+using System.Collections.Generic;
+
+namespace ACadSharp.IO.Templates
+{
+	internal class CadSplineTemplate : CadEntityTemplate
+	{
+        private CSMath.XYZ _currentControlPoint = new CSMath.XYZ();
+        private CSMath.XYZ _currentFitPoint = new CSMath.XYZ();
+
+        public CadSplineTemplate(Spline entity) : base(entity) { }
+
+        public override bool CheckDxfCode(int dxfcode, object value)
+        {
+            bool found = base.CheckDxfCode(dxfcode, value);
+            if (found)
+                return found;
+
+            var spline = this.CadObject as Spline;
+
+            switch(dxfcode)
+            {
+                //--- Knot
+                case 40:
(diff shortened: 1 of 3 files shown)


ACadSharp/IO/Templates/CadSplineTemplate.cs
```diff
@@ -0,0 +1,105 @@
...
+                case 41:
+                    spline.Weights.Add((double)value);
+                    return true;
+
+                //--- ControlPoint
+                case 10:
+                    this._currentControlPoint = new CSMath.XYZ(
+                        (double)value,
+                        _currentControlPoint.Y,
+                        _currentControlPoint.Z
```
DomCR, line 39: PR looks really good, but in this point I've spotted a reading error:
When this code is read, this should reset the field ```_currentControlPoint``` like ```new CSMath.XYZ((double)value, 0, 0)``` the way is written right now, at the second point it will read the Y and Z from the first one.

ACadSharp/IO/Templates/CadSplineTemplate.cs
```diff
@@ -0,0 +1,105 @@
...
+                        );
+                    spline.ControlPoints.Add(this._currentControlPoint);
+                    return true;
+
+                //--- FitPoint
+                case 11:
+                    this._currentFitPoint = new CSMath.XYZ(
+                        (double)value,
+                        _currentFitPoint.Y,
+                        _currentFitPoint.Z
```
DomCR, line 63: For the FitPoints it works the same way as the ControlPoints

DomCR (changes requested):

DomCR:
Thanks a lot for the PR, please check the changes that I propose to fix the potential reading error and I'll merge it.

DomCR:
Also, for the Lists of elements in dxf there is an attribute with a ```DxfReferenceType.Count``` as parameter, but is not implemented yet

ACadSharp/IO/Templates/CadSplineTemplate.cs
michal-kindl: Thank you for correction, I do check the proposed changes.

DomCR (approved):

PR #43, opened 2022-07-23
issue42 ellipse fix

DomCR:
# Description

- Fix ellipse reading for the Endpoint field
- Updated documentation


mattbrownfaro:
Update: I made the fix at my end. Instead of reading the point into EndPoint, it was loading it into a local variable.

Change made here:
DwgObjectSectionReader.cs
readEllipse()
Line 2222
			ellipse.EndPoint = this._objectReader.Read3BitDouble();

PR #46, opened 2022-07-27
Reading & Writing Performance Review Changes

DJGosnell:
Upon my first review, I have found that reflection and list access are the immediate worst offenders at this point.  I ran a read & write operation as indicated below and found the following with a CAD DXF drawing that has roughly 22,000 items (Blocks & simple entities).  I would attach the DXF, but it is a private project file.

For this first pass, I tired to leave the base code alone as much as possible as my initial work on refactoring the Map classes to use static instances became wide reaching very fast.

With the current changes, we get a performance increase of roughly 5x on the first read and then 6.5x faster on subsequent reads over the baseline. The write performance is also now 4.7x faster.

Ran into a small reading issue with `DxfSectionReaderBase.readMapped` method where it was trying to read a `DxfReferenceType.Handle` for a FontStyle but the file actually contains a string value.  I'll address that in a separate PR.

#### Changes
- Reflection used for setting properties has now been changed out for using Expressions for generating cached setters & getters for properties via the `PropertyExpression` class.
- CadObjectCollection<T>.Entities is now a `HashMap<T>` due to `Contains()` lookups during addition of entities.
- `DxfMap` and `DxfClassMap` now both cache their results of the `Create<T>()` method.  Caches are cleared via the `ClearCache()` method on both types.  Caching these items reduces the reflection on the first read/write significantly and all future reads and writes have all the maps already cached and are thus faster.

``` csharp
var sw = Stopwatch.StartNew();
Console.WriteLine("Startup Read & Write");

var document = DxfReader.Read("X-Base.dxf", (o, e) => { });
Console.WriteLine($"[{sw.ElapsedMilliseconds:0,000}ms] Completed Read");
sw.Restart();
DxfWriter.Write("X-Base-output.dxf", document, false);
Console.WriteLine($"[{sw.ElapsedMilliseconds:0,000}ms] Completed Write");

Console.WriteLine($"Completed startup Read & Write.");
Console.WriteLine();

sw.Restart();
for (int i = 0; i < 1; i++)
{
    sw.Restart();
    Console.WriteLine($"Started Subsequent Read & Write number {i + 1}");
    document = DxfReader.Read("X-Base.dxf", (o, e) => { });
    Console.WriteLine($"[{sw.ElapsedMilliseconds:0,000}ms] Completed Read");

    sw.Restart();
    DxfWriter.Write("X-Base-output.dxf", document, false);
    Console.WriteLine($"[{sw.ElapsedMilliseconds:0,000}ms] Completed Write");

    Console.WriteLine($"Completed Subsequent Read & Write number {i + 1}");
    Console.WriteLine();
}
```

## Baseline Performance
``` 
Startup Read & Write
[5,693ms] Completed Read
[2,264ms] Completed Write
Completed startup Read & Write.

Started Subsequent Read & Write number 1
[5,534ms] Completed Read
[2,278ms] Completed Write
Completed Subsequent Read & Write number 1
```
#### Dot Trace Performance
image

## After PR Modifications
```
Startup Read & Write
[1,114ms] Completed Read
[0,486ms] Completed Write
Completed startup Read & Write.

Started Subsequent Read & Write number 1
[0,872ms] Completed Read
[0,495ms] Completed Write
Completed Subsequent Read & Write number 1
```
#### Dot Trace Performance
image

### DIFF
--- a/ACadSharp.Tests/Tables/BlockRecordTests.cs
+++ b/ACadSharp.Tests/Tables/BlockRecordTests.cs
@@ -3,6 +3,7 @@ using ACadSharp.Tables;
 using ACadSharp.Tests.Common;
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text;
 using Xunit;
 
@@ -78,9 +79,15 @@ namespace ACadSharp.Tables.Tests
 
 			CadObjectTestUtils.AssertTableEntryClone(record, clone);
 
+			// Copy the state of the entities to an array as this is now using a HashMap for performance
+			// and cannot be accessed via indexes.
+
+            var recordEntities = record.Entities.ToArray();
+            var cloneEntities = clone.Entities.ToArray();
+
 			for (int i = 0; i < record.Entities.Count; i++)
 			{
-				CadObjectTestUtils.AssertEntityClone(record.Entities[i], clone.Entities[i], true);
+				CadObjectTestUtils.AssertEntityClone(recordEntities[i], cloneEntities[i], true);
 			}
 		}
 	}
--- a/ACadSharp/CadObjectCollection.cs
+++ b/ACadSharp/CadObjectCollection.cs
@@ -12,13 +12,14 @@ namespace ACadSharp
 
 		public event EventHandler<ReferenceChangedEventArgs> OnRemove;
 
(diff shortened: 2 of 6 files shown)


DomCR (approved):
Really great work!

Issue #80, opened 2022-12-08
Angle Differences DXF & DWG


DJGosnell:
While working on DXF & DWG files in which the content is the same, I've noticed that the DWG reader reads angle values in radians (as documented)  I have a branch which shows these differences in MText, TextEntities, Arcs & Ellipses.

TextEntities show degrees on the DXF & Arc start and end angles.  All others are radians.  I was just going to make a PR, but I thought it would be worth discussing.  the DWG documentation rightfully states that the angle values will be in radians but the DXF is a little less explicit by stating "angle" for several types.

[netDXF](https://github.com/haplokuon/netDxf) handles this by ensuring all the angles are in degrees, but it seems that most of the angles in the files are actually in radians.  Do we want to just convert everything to make sure it's in radians during the reading process?

Angle DXF & DWG written in AutoCAD 2010 formats.
[angles.zip](https://github.com/DomCR/ACadSharp/files/10190564/angles.zip)

A side note is that the MText rotation is not being read by either.  I'll open a separate issue for this.

Output of the [dwg-dxf-angles branch](https://github.com/DJGosnell/ACadSharp/tree/dwg-dxf-angles):
```
DXF TextEntity Rotation: 45
DWG TextEntity Rotation: 0.7853981633974483
DXF Ellipse StartParameter: 1.020388469764473
DWG Ellipse StartParameter: 1.0203884697644738
DXF Ellipse StartParameter: 1.570796326794896
DWG Ellipse StartParameter: 1.5707963267948966
DXF Arc StartAngle: 45
DWG Arc StartAngle: 0.7853981633974483
DXF Arc EndAngle: 90
DWG Arc EndAngle: 1.5707963267948966
DXF MText Rotation: 0
DWG MText Rotation: 0
```

![image](https://user-images.githubusercontent.com/224169/206588101-1ec33875-a218-44dc-968e-7fc0e71551ba.png)

DomCR:
I'll keep them as **radians**, C# works with them so it's easier to work with `System.Math` functions if the angles are in radians.

DomCR:
My priority right now is to finish the first alpha for the `DwgWriter` I can open a branch dedicated to this issue but I'll not addressed for a while.

DJGosnell:
Excellent.  I'll issue a PR for this.

PR #83, opened 2022-12-09
DXF Angle Reading Conversion

DJGosnell:
It was discovered that angles read from `Arc`, `Insert` and `TextEntity` entities were in degrees while other DXF angles are in radians and all DWG angles are in radians.  This PR adds a new CadArcTemplate and modifies the  CadInsertTemplate and CadTextEntityTemplate to handle the conversion while reading DXF files.

I went ahead and added the new CadArcTemplate to DwgObjectSectionReader for consistency sake, even though it does not do anything at this time over the base CadEntityTemplate.

### DIFF
--- a/ACadSharp/IO/Templates/CadArcTemplate.cs
+++ b/ACadSharp/IO/Templates/CadArcTemplate.cs
@@ -1,0 +1,25 @@
+﻿using System;
+using ACadSharp.Entities;
+using ACadSharp.IO.DXF;
+using ACadSharp.Tables;
+
+namespace ACadSharp.IO.Templates
+{
+	internal class CadArcTemplate : CadEntityTemplate
+	{
+
+		public CadArcTemplate(Entity entity) : base(entity) { }
+
+
+        public override void Build(CadDocumentBuilder builder)
+		{
+			base.Build(builder);
+
+            if (builder is DxfDocumentBuilder && this.CadObject is Arc arc)
+            {
+                arc.StartAngle *= MathUtils.DegToRad;
+                arc.EndAngle *= MathUtils.DegToRad;
+            }
+		}
+	}
+}
--- a/ACadSharp/IO/Templates/CadInsertTemplate.cs
+++ b/ACadSharp/IO/Templates/CadInsertTemplate.cs
@@ -1,6 +1,7 @@
 ﻿using ACadSharp.Entities;
 using ACadSharp.Tables;
 using System.Collections.Generic;
+using ACadSharp.IO.DXF;
 
 namespace ACadSharp.IO.Templates
 {
@@ -47,7 +48,8 @@ namespace ACadSharp.IO.Templates
 		{
 			base.Build(builder);
 
-			Insert insert = this.CadObject as Insert;
+            if (!(this.CadObject is Insert insert))
+                return;
 
 			if (builder.TryGetCadObject(this.BlockHeaderHandle, out BlockRecord block))
 			{
@@ -72,6 +74,11 @@ namespace ACadSharp.IO.Templates
 			{
 				insert.Attributes.Seqend = seqend;
 			}
+
+            if (builder is DxfDocumentBuilder)
+            {
+                insert.Rotation *= MathUtils.DegToRad;
+            }
 		}
 	}
 }
--- a/ACadSharp/IO/Templates/CadTextEntityTemplate.cs
+++ b/ACadSharp/IO/Templates/CadTextEntityTemplate.cs
@@ -1,4 +1,6 @@
-﻿using ACadSharp.Entities;
+﻿using System;
+using ACadSharp.Entities;
+using ACadSharp.IO.DXF;
 using ACadSharp.Tables;
 
 namespace ACadSharp.IO.Templates
@@ -21,7 +23,7 @@ namespace ACadSharp.IO.Templates
 			}
 		}
 
-		public override void Build(CadDocumentBuilder builder)
+        public override void Build(CadDocumentBuilder builder)
 		{
 			base.Build(builder);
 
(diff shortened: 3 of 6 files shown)


DJGosnell:
Side note, I noticed that `Point.Rotation` is a property but how you rotate a point is beyond me.  Not sure about `DimensionLinear.Rotation` either as the dimensions are aligned explicitly along two points. I'll look at that later.

DomCR:
> It was discovered that angles read from `Arc`, `Insert` and `TextEntity` entities were in degrees while other DXF angles are in radians and all DWG angles are in radians. This PR adds a new CadArcTemplate and modifies the CadInsertTemplate and CadTextEntityTemplate to handle the conversion while reading DXF files.
> 
> I went ahead and added the new CadArcTemplate to DwgObjectSectionReader for consistency sake, even though it does not do anything at this time over the base CadEntityTemplate.
> 
> Refer to #80 for more details.

I'll check why you cannot point to my branch, by now I redirected your implementation.

ACadSharp/IO/Templates/CadInsertTemplate.cs
```diff
@@ -72,6 +74,11 @@ public override void Build(CadDocumentBuilder builder)
 			{
 				insert.Attributes.Seqend = seqend;
 			}
+
+            if (builder is DxfDocumentBuilder)
+            {
+                insert.Rotation *= MathUtils.DegToRad;
```
DomCR, line 80: Instead of checking if the builder is a `DxfDocumentBuilder` I would prefer to override the method:
```CSharp
public virtual bool CheckDxfCode(int dxfcode, object value)
```
This method bypassed the mapper when the DxfReader assigns the values, this way we'll keep the assignation in the dxf scope.

ACadSharp/IO/Templates/CadTextEntityTemplate.cs
```diff
@@ -21,14 +23,20 @@ public override bool CheckDxfCode(int dxfcode, object value)
...
 
 			switch (this.CadObject)
 			{
 				case TextEntity text:
 					text.Style = builder.GetCadObject<TextStyle>(this.StyleHandle);
+
+                    // When the rotation is read in a DXF, the value is in decimal, but when the value
+                    // is read in a DWG, it is in radians.  Convert only on DXFs. Issue #80
+                    if (builder is DxfDocumentBuilder)
+                        text.Rotation *= MathUtils.DegToRad;
```
DomCR, line 38: Same case:
```CSharp
public virtual bool CheckDxfCode(int dxfcode, object value)
```

ACadSharp/IO/Templates/CadArcTemplate.cs
```diff
@@ -0,0 +1,25 @@
...
+
+
+        public override void Build(CadDocumentBuilder builder)
+		{
+			base.Build(builder);
+
+            if (builder is DxfDocumentBuilder && this.CadObject is Arc arc)
+            {
+                arc.StartAngle *= MathUtils.DegToRad;
+                arc.EndAngle *= MathUtils.DegToRad;
```
DomCR, line 21: Instead of checking if the builder is a `DxfDocumentBuilder` I would prefer to override the method:
```CSharp
public virtual bool CheckDxfCode(int dxfcode, object value)
```
This method bypassed the mapper when the DxfReader assigns the values, this way we'll keep the assignation in the dxf scope.

ACadSharp/IO/Templates/CadInsertTemplate.cs
DomCR, line 80: Same case:
```CSharp
public virtual bool CheckDxfCode(int dxfcode, object value)
```

ACadSharp/IO/Templates/CadTextEntityTemplate.cs
DomCR, line 38: Same case:
```CSharp
public virtual bool CheckDxfCode(int dxfcode, object value)
```

DomCR (changes requested):

DomCR:
About the rotation I can take a look, but the operation should go like this:

**In 2D** is quite easy, just take the `XY` from the `InsertPoint` and create the new vector, `AlignmentPoint`, using the `Rotation` angle that is being set, in the other case, you just need to calculate the angle between the `XY` from the `InsertPoint` and the `XY` from `AlignmentPoint`.

**In 3D** we will need to calculate the plane that `InsertPoint` and the `Normal` creates, where the Text normal is also the normal of the plane, and then find the `AlignmentPoint` in that plane, create the 2D points equivalent in this plane and make the calculations.

You said that this is something that you need done ASAP, so once the rotation for 2D is done this can be merge and we can take care for the 3D rotation in another issue.

DomCR:
Here is an example of how to override the `CheckDxfCode` method in `CadLayerTemplate`

```CSharp
public override bool CheckDxfCode(int dxfcode, object value)
{
	bool found = base.CheckDxfCode(dxfcode, value);
	if (found)
		return found;

	switch (dxfcode)
	{
		case 348:
			// Code not documented
			found = true;
			break;
	}

	return found;
}
```

DomCR:
> It looks like `CheckDxfCode` would be the place to make these modifications, but it appears that the `CheckDxfCode` methods are only called on codes that are not bound by the `DxfCodeValueAttribute`. When I remove the `DxfCodeValueAttribute` for the `TextEntity.Rotation`, the the CheckDxfCode is called, but otherwise it is not even invoked on the DXF I have attached.
> 
> Is there another location that is specific to the DXF parsing path that these angle values could be updated or am I just missing something?
> 
> [angle.zip](https://github.com/DomCR/ACadSharp/files/10212633/angle.zip)

You are right... it only jumps the `readMapped` if the attribute is missing or is a `ControlString`.

I'll merge the PR as it is and open a new issue to refactor this part of the code, the Dxf reader is quite delicate and I don't want to block this.

PR #84, opened 2022-12-12
Extends AutoCAD Testing Versions

DJGosnell:
This PR adds a lookup for AutoCAD installations if it does not exist in the default location.  Allows testing on version of AutoCAD from 2023 to 2021 LT and full.


DomCR:
Good work!! Thanks!

PR #87, opened 2022-12-14
Fix for TextVerticalAlignmentType

DJGosnell:
Correction for vertical justification order.
image


DJGosnell:
I've noticed that there are multiple files with differing indention styles.  Some tabs, some spaces.  When I propose an update like this, would you prefer that the files have indention consistent with the `.editorconfig` spaces or just leave the indention style of the file to make the PR more simple?

DomCR:
The `.editorconfig` needs to be updated and configured properly.

PR #88, opened 2022-12-14
Revert Text/MText Linking of Rotation and AlignmentPoint

DJGosnell:
While testing these changes in production, it was found that since the MText.Rotation set the AlignmentPoint and vice versa, the former would break the correct alignment points which are required for correct text placement.

We need to review how to handle this properly.  Right now, I think the best course of action would be to unlink the two and allow for the library consumer to determine how to properly handle the values.  I am not even sure that it is actually possible to convert from a `Rotation` radian to a 3D `AlignmentPoint`.

### DIFF
--- a/ACadSharp.Tests/Entities/MTextTests.cs
+++ b/ACadSharp.Tests/Entities/MTextTests.cs
@@ -1,0 +1,24 @@
+﻿using System;
+using ACadSharp.Entities;
+using CSMath;
+using System.Collections.Generic;
+using System.Linq;
+using Xunit;
+
+namespace ACadSharp.Tests.Entities
+{
+    public class MTextTests
+    {
+
+        [Fact]
+        public void RotationDoesNotOverrideAlignmentPoint()
+        {
+            var text = new MText
+            {
+                Rotation = Math.PI
+            };
+
+            Assert.Equal(text.AlignmentPoint, XYZ.Zero);
+        }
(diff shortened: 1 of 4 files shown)


DJGosnell:
Side note, do you have any more direct contact you use?  Email, Discord, etc.?  I'd love to be able to talk about certain parts of this project more in a conversational environment if you would be willing.

DomCR:
I've opened the discussions in Github

ACadSharp.Tests/Entities/MTextTests.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        [Fact]
+        public void RotationDoesNotOverrideAlignmentPoint()
+        {
+            var text = new MText
+            {
+                Rotation = Math.PI
+            };
+
+            Assert.Equal(text.AlignmentPoint, XYZ.Zero);
```
DomCR, line 21: This test needs to be reviewed, `AlignmentPoint` is a unitary vector, that means that it will never be 0, the modulus will always be 1 (NOTE: this has been implemented as a point so the modulus can be higher by now)

Also, PI in radians is 180 degrees, meaning the alignment point will be (-1,0,0), right now in the test the result is (-1,1.2246467991473532E-16,0) which is "correct" but it needs to correct the almost zero to 0.

PR #93, opened 2023-01-02
True Color Handling

DJGosnell:
### Background and motivation
The `Color` struct does not currently handle true color storage.  This PR adds that functionality.

The true color needs to be used for multiple types of entities and adjusting the Color struct seemed to be the most simple way to add that functionality.   This PR came out of the need of true colors in MText values.

### Changes
Fundamentally changes how colors are stored in the struct.
Structure is now set to `readonly` as colors it complicates the storage of the true color and indexed color.  Also provides the CLR with potential opportunities to optimize things.

Colors are now stored in an integer in the `_color` field.  The integer has the 30th bit of the `_color` field set to 1 if the value stored is a true color or the 30th bit is set to 0 if the value is an indexed color.  Handling color storage this way has the benefit of preventing us from having to store a completely unused indexed color when the stored value is a true color and vice versa.  Also prevents having to keep a bool stored keeping track if the value stored is true color.

I added the static `Color.FromTrueColor` method to prevent usage confusion.  The CLR will normally default to passing a number value as an integer instead of a short.  Using a static method to make the struct into a true color prevents the potential confusion in which constructor is used.

### Old Contract
```csharp
public struct Color { 
  bool IsByBlock { get; }
  bool IsByLayer { get; }
  short Index { get; set; }
}
```

### New Contract Changes
```csharp
public readonly struct Color { 
  bool IsByBlock { get; }
  bool IsByLayer { get; }
  short Index { get; }
  int TrueColor { get; }
  bool IsTrueColor { get; }
}
```

### Misc other changes.
- Changed `Color.IndexRgb` to be a static array.  The index of the array aligns with the True Color approximation.


DJGosnell:
Going to check on the testing failures tomorrow.

ACadSharp.Tests/ACadSharp.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFrameworks>net48;net5.0;net6.0</TargetFrameworks>
```
DomCR, line 4: I would suggest to change this in another branch so we don't break the scope for this task.

ACadSharp/Color.cs
```diff
@@ -5,297 +5,445 @@
...
-			{248, new byte[] {38, 0, 9}},
-			{249, new byte[] {88, 19, 23}},
-			{250, new byte[] {0, 0, 0}},
-			{251, new byte[] {101, 101, 101}},
-			{252, new byte[] {102, 102, 102}},
-			{253, new byte[] {153, 153, 153}},
-			{254, new byte[] {204, 204, 204}},
-			{255, new byte[] {255, 255, 255}}
-		};
+    public readonly struct Color
```
DomCR, line 11: A are we sure that we want to use a `readonly struct`?? This may simplify the implementation for the development side but lacks flexibility to mutate the struct.

I'm not against I just want to know the reason for this decision.

ACadSharp/Color.cs
```diff
@@ -5,297 +5,445 @@
...
+        {
+            if (IsTrueColor)
+            {
+                return new ReadOnlySpan<byte>(BitConverter.GetBytes(_color), 0, 3);
+            }
+
+            return default;
+        }
+
+        public bool Equals(Color other)
```
DomCR, line 414: Please add the interface `System.IEquatable<Color>` so the different `System` libraries can use this method.

ACadSharp.Tests/ACadSharp.Tests.csproj
DJGosnell: Agreed. This was done mostly for #94 anyway. Not sure why I included that here. It ended up complicating things on the CI pipeline anyway.

ACadSharp/Color.cs
DJGosnell: Will do!

ACadSharp/Color.cs
DJGosnell: I have to dig again to find some of the articles, but essentially having mutable structs can have unintended consequences during usages.  My thoughts are if this is headed off at the pass with a `readonly struct` then headaches can be avoided later. See the two articles below.
https://stackoverflow.com/a/441323
https://learn.microsoft.com/en-us/archive/blogs/ericlippert/mutating-readonly-structs

I'm honestly open to changing it back to the way you originally had it since it is effectively a struct acting as a simple wrapper for an int32. I was just trying to avoid potential foot guns for later.

I'll await your decision.

DomCR: Keep it this way, as I said, I know that this makes implementing way easier, but we have to keep in mind that Autocad Color also has the ColorName field, check the DxfCode 430, but is not tested nor implemented, but I'm guessing that is for colors that are not in the dictionary so the RGB values may need to be open for change in future implementations.

DJGosnell: This is a good point.  I didn't even know was a thing in the design.  I've been working with DWGs and DXFs for a while but I've not had a need of a named color before other than the pseudo named red, green, blue colors that ACAD translates to indexed colors.  I'll try to run through my bank of DWGs and see if any of them use this code.

ACadSharp.Tests/ACadSharp.Tests.csproj
```diff
@@ -12,6 +12,7 @@
     <PackageReference Include="MSTest.TestAdapter" Version="2.1.1" />
     <PackageReference Include="MSTest.TestFramework" Version="2.1.1" />
     <PackageReference Include="coverlet.collector" Version="1.3.0" />
+    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
```
DomCR, line 15: This package is not needed if we are using `netcoreapp3.1`

ACadSharp.Tests/ACadSharp.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+	<TargetFramework>netcoreapp3.1</TargetFramework>
```
DomCR, line 4: Can you match the indentation? I'll look for a way to configure it in the .editorconfig

ACadSharp.Tests/Common/DocumentIntegrity.cs
```diff
@@ -88,9 +89,12 @@ public void AssertBlockRecords(CadDocument doc)
 		public void AssertDocumentTree(CadDocument doc)
 		{
 			this._document = doc;
+#if NETFRAMEWORK
```
DomCR, line 92: Same comment as : https://github.com/DomCR/ACadSharp/pull/93/files#r1067939549

ACadSharp/ACadSharp.csproj
```diff
@@ -32,4 +32,8 @@
     </None>
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="System.Memory" Version="4.5.5" />
```
DomCR, line 36: Is this package needed? I see that you are using `ReadOnlySpan` to return the bytes for the `GetTrueColorRgb` and `GetIndexRGB` but is there any difference if we return an array of bytes?

DomCR:
I've updated the .editorconfig in master, please check if it solves the indentation problems.

Thanks!

ACadSharp.Tests/ACadSharp.Tests.csproj
DJGosnell: Yes.  I accidentally replaced it with a tab...

ACadSharp.Tests/ACadSharp.Tests.csproj
DJGosnell: Once again, this is was an items copied from #94 by accident.  Will fix.

ACadSharp/ACadSharp.csproj
DJGosnell: This package is implicitly included on all .NET Core projects by default as it is part of core packages.  This will only be included on the .NET Framework 4.8 and ignored by all the other targets. So effectively, it is not adding a package requirement except for the 4.8 target.

In the `GetTrueColorRgb` method, I'm using the BitConverter.GetBytes(_color) to convert the int32 into four bytes representing the individual RGB color values.  The fourth byte is used by the struct to determine if the color is a true color or indexed color.  For the end-user the fourth byte is an implementation detail they don't and shouldn't care about.  Using a ReadOnlySpan<byte> to slice this array to three bytes prevents a new allocation and copy of a new 3 byte array from the 4 byte array.

In the `GetIndexRGB` method, I'm calling the .AsSpan(); method on the `_indexRgb` byte array which allows for a reference to the contents but prevents someone from modifying the values them (by accident or otherwise).

The `System.Memory` package is also required by PR #94 extensively which allows for a zero copy reading and writing of MText Values.  I do not see this as adding anything too heavy to ACadSharp.  Trust me, I'm the first one to try to remove packages and reduce package bloat, but this is a very small concession to allow 4.8 users to use Spans for increased efficiency.

ACadSharp.Tests/ACadSharp.Tests.csproj
DJGosnell: Just confirming, the .editorconfig file has the project using tabs `indent_style = tab`.  I'm showing tabs, but the original file was spaces.  Can you please confirm tabs are what is intended?

DomCR: Correct, that's the right configuration.

ACadSharp/ACadSharp.csproj
DomCR: Thanks for the clarification! I see that we are on the same page about packages.

DomCR:
Please revisit the comment on https://github.com/DomCR/ACadSharp/pull/93#discussion_r1067940703 to remove the pre-compiler directive due it's not necessary and the `using Newtonsoft.Json` and this is good to go.

Great stuff!

DJGosnell:
There.  Should be good now.

DomCR (approved):

PR #125, opened 2023-05-05
Added Get function for ExtendedDataCollection

sraedler:
# Description

I added Dictionary get functions to ExtendedDataCollection because I need them in a project.
IMHO it would also be possible to inherit the ExtendedDataCollection from Dictionary, but I was not willing to change this because it might have an unforeseen impact on other code parts.

# Tasks done in this PR
* [x] Something awesome.
* [ ] An evil bug have been defeted.

# Notes for reviewer
 - This is just a small contribution and it is just because I require it. If it is not of interest for the project at all, please reject.

Thanks for your effort. This project increased my parsing performance of DWG file by factor 5 compared to http://www.ivancad.com/

### DIFF
--- a/ACadSharp/ExtendedDataCollection.cs
+++ b/ACadSharp/ExtendedDataCollection.cs
@@ -14,6 +14,21 @@ namespace ACadSharp
 			this._data.Add(app, edata);
 		}
 
+		public ExtendedData Get(AppId app)
+		{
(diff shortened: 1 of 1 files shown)


DomCR:
Hi @sraedler,

thanks for the contribution! the only think that I would ask is to document the methods and I'll merge it.

PR #201, opened 2023-11-09
Issue 198 index out of range in readmultileader

DomCR:
# Description

Fix for index out of range in readmultileader in `DwgObjectReader`

# TODO

Consider to implement the transformation matrix in multileader context using `CSMath.Matrix4`


mme1950:
Thank you for fixing our errors.

Issue #86, opened 2022-12-14
3D rotation for Arc, Insert and TextEntity


DomCR:
`Arc`, `Insert` and `TextEntity` have a rotation that is calculated in the set in case that the `AlignmentPoint` or the `Rotation` changes.

We need an investigation to make sure that this only affects the 2D, X and Y, of the coordinates and does not use the Normal to calculate the relative plane.

If that's the case, the methods will need to change to calculate the angle in 3D using the Normal.

PR #82, opened 2022-12-09
MText refactoring

DomCR:
# Description

- Documentation update and fix
- Rotation and AlignmentPoint linked to be coherent 
- Fix for the Rotation value not being set - Only for 2D

Issue #81, opened 2022-12-08
MText Rotation Not Being Read


DJGosnell:
In issue #80, I ran across the issue with the rotation not being read on MText entities on both the DWG and DXF readers.  See the sample file below [here](https://github.com/DomCR/ACadSharp/files/10190564/angles.zip).

I'll start to dig into this when I have some time.

DomCR:
In the Dwg Reader there is this line:
```CSharp
//X-axis dir 3BD 11 Apparently the text x-axis vector. (Why not just a rotation?) ACAD maintains it as a unit vector.
mtext.AlignmentPoint = this._objectReader.Read3BitDouble();
```

So I'm guessing that the rotation is calculated using the `AlignmentPoint` and the `InsertPoint`.

For the Dxf we need to detect if the angle is provided or not, and then calculate the `AlignmentPoint` (which it should be renamed as a vector) or the angle depending on which one is provided.

About **how to save the angles**, I'll keep them as **radians**, C# works with them so it's easier to work with `System.Math` functions if the angles are in radians.

DJGosnell:
That is exactly the information that I need. Confirmed that the `AligmentPoint` and `InsertPoint` were both properly set to determine the angle.

Since the angle is determined If this is the case, is there any time that `MText.Rotation` is actually used?  I see it has `DxfCodeValue(50)` attribute so I'm presuming that it is read at some point, but I feel like whoever uses this library needs to determine which value to use.

In this case, do we add any "helper" methods (or property) like `MText.GetRotationAngle` and `MText.SetRotationAngle` on the class to ease the use of this or do we just leave it up to the Libraryconsumer to determine which to use (`AligmentPoint` and `InsertPoint` vs `Rotation`)?

I'm willing to do this if I have a better understanding if/when the `Rotation` DXF code is used.

DomCR:
It seems that the DWG works only with the `AlignmentPoint` because checking the documentation the rotation is not read at any point, so for DWG when `AlignmentPoint` is set, the Rotation automatically should get it's value.

For DXF it seems that you have both options, but I presume that both values must be coherent with each other, if not, Autocad just picks the last value read:
[...]

PR #203, opened 2023-11-14
Text rotation fix

DJGosnell:
# Description

Removed internal linking of the `TextEntity.Rotation` and `TextEntity.AlignmentPoint` properties.

Fixes issues with rotation adjusting AlignmentPoint and AlignmentPoint adjusting the rotations.  After a bit of research, the AutoCad AlignmentPoint is the location where the text is to be inserted in conjunction with the VerticalAlignment and HorizontalAlignment.  Based upon certain configurations of the enumns, either the InsertPoint or the AlignmentPoint is used to determine where the text is to be inserted.  As far as I can tell, it does not have anything to do with the Rotation, at least as far as 2D goes.  I do not have time to experiment with 3d rotations. Ref #86.

For reference, the `InsertionPoint`s below are the bottom left grips while the `AlignmentPoint`s are the other grips on the text.
image

#81 
#82


DomCR:
We have a similar implementation for `MText`, do you know if this works the same for it?

PR #207, opened 2023-11-20
ReadEnColor fix

DomCR:
# Description

`ReadEnColor` fix for true color


mme1950:
@DomCR This PR seems to be still open.

Issue #151, opened 2023-07-17
Dimension Text default logic and formatting


mme1950:
The `Dimension.Text` property should return an empty or explicit text when the value is not null.
It should return a default text, i. e. the formatted Measurement, when the value is null.
The implementation handles null and empty in the same way.

<pre>
		/// <summary>
		/// Dimension text explicitly entered by the user
		/// </summary>
		/// <remarks>
		/// Optional; default is the measurement.
		/// If null, the dimension measurement is drawn as the text, 
		/// if ““ (one blank space), the text is suppressed.Anything else is drawn as the text
		/// </remarks>
		[DxfCodeValue(DxfReferenceType.Optional, 1)]
		public string Text
		{
			get { return string.IsNullOrEmpty(_text) ? this.Measurement.ToString() : this._text; }
			set { this._text = value; }
		}
</pre>

I would propose to format the `Measurement` value after rounding to `Dimension.Style.DecimalPlaces` and using the `Dimension.Style.DecimalSeparator`. Perhaps it even makes sense to add unit string etc.

mme1950:
[This documentation](https://help.autodesk.com/view/OARX/2023/ENU/?guid=GUID-EDD54EAC-A339-4EBA-AEA6-EC8066505E2B) contains the description for group code 1 as given in the comment above.

_Dimension text explicitly entered by the user. Optional; default is the measurement. If null or “<>”, the dimension measurement is drawn as the text, if ““ (one blank space), the text is suppressed. Anything else is drawn as the text_.

To suppress the text a **blank** has to be entered. Null or an empty sting cannot be distinguished.

An explicit text appears as entered. All other settings such as prefix, suffix, alternate measurement including alternate prefix and suffix, tolerance etc. are ignored.

But I think it would be nice to include all these settings in the default `Text` property.

DomCR:
Perfect! I'll take a look to implement this changes.

Thanks!

mme1950:
Hi @DomCR
including prefix and suffix, tolerance etc. requires some investigation.

This code would be half way. Shall I create PR?

		public string Text
		{
			get {
				if (this._text == " ")
				{
					return string.Empty;
				}

				if (string.IsNullOrEmpty(this._text))
				{
					double measurement = Math.Round(this.Measurement, this.Style.DecimalPlaces);
					NumberFormatInfo nfi = new NumberFormatInfo();
					nfi.NumberDecimalSeparator = this.Style.DecimalSeparator.ToString();
[...]

PR #223, opened 2023-12-06
20231206 nka 151 dimension text default logic and formatting

nka1994:
# Description

Short description about this PR

# Tasks done in this PR
* [ ] Something awesome.
* [ ] An evil bug have been defeted.

# Notes for reviewer

### DIFF
--- a/ACadSharp/Entities/Dimension.cs
+++ b/ACadSharp/Entities/Dimension.cs
@@ -155,19 +155,16 @@ namespace ACadSharp.Entities
 		public bool FlipArrow2 { get; set; }
 
 		/// <summary>
-		/// Dimension text explicitly entered by the user
+		/// Gets or sets an explicit dimension text to be displayed instead of the standard
+		/// dimension text created from the measurement in the format specified by the
+		/// dimension-style properties.
 		/// </summary>
 		/// <remarks>
-		/// Optional; default is the measurement.
-		/// If null, the dimension measurement is drawn as the text, 
-		/// if ““ (one blank space), the text is suppressed.Anything else is drawn as the text
+		/// If null or empty, the dimension created from the measurement is to be displayed. 
+		/// If " " (one blank space), the text is to be suppressed. Anything else is drawn as the text.
 		/// </remarks>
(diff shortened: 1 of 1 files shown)


ACadSharp/Entities/Dimension.cs
```diff
@@ -106,18 +107,36 @@ public DimensionType Flags
...
+		/// If null or empty, the formatted <see cref="Measurement"/> value is returned, 
+		/// if ““ (one blank space), the text is suppressed. Anything else is drawn as the text.
+		/// </value>
 		[DxfCodeValue(DxfReferenceType.Optional, 1)]
 		public string Text
 		{
-			get { return string.IsNullOrEmpty(_text) ? this.Measurement.ToString() : this._text; }
-			set { this._text = value; }
+			get {
+				if (this._text == " ")
```
DomCR, line 121: I would put this if after the `IsNullOrEmpty` and then check the _text using the method `string.IsNullOrWhiteSpace` due this does not check for a string with multiple white spaces.

mme1950: ... Anything else is drawn as the text. This includes multiple white spaces. Hmm - not sensible.

anyhow - after some work with dimensions I think my proposal is not a good idea, since the generated text may be incomplete or not suitable for angular dimensions. And there is no indication, whether the text was created according to the various properties or explicitly entered. Please see #223.

mme1950:
Hi @DomCR,

after some work with dimensions I think my proposal is not a good idea, since the generated text may be incomplete or not suitable for angular dimensions. And there is no indication, whether the text was created according to the various properties or explicitly entered. The implementation is definitely better than the existing one but works well only for linear dimensions with no special options.

The generated text must consider all settings specified in DimensionStyle including prefixes, postfixes, alternate-unit dimensioning, and tolerance. All properties may be overriden in the entity's extended data. The implementation for linear and angular dimensions must be different.

This would require implemantation of functions to read extended properties. Should this be provided by ACadSharp?

mme1950:
please see #151.

mme1950:
Please close this PR. I had trouble to update this branch, I will create a new PR.

DomCR (approved):

Issue #226, opened 2023-12-11
DimensionStyle: DimArrow1 and DimArrow2 must have Type BlockRecord


mme1950:
The properties `DimArrow1` and `DimArrow2` of the DimensionStyle class provide alternate arrowheads for a dimension line. Both properties currently are typed as `Block` but must be `BlockRecord`.

Presumably also the properties `LeaderArrow` and `ArrowBlock` must be of type `BlockRecord`.

PR #229, opened 2023-12-15
#226

mme1950:
-  LeaderArrow, ArrowBlock, DimArrow1, DimArrow2: type changed to BlockRecord
-  Comments added from https://help.autodesk.com/view/ACDLT/2024/

### DIFF
--- a/ACadSharp/IO/Templates/CadDimensionStyleTemplate.cs
+++ b/ACadSharp/IO/Templates/CadDimensionStyleTemplate.cs
@@ -32,17 +32,17 @@ namespace ACadSharp.IO.Templates
 				this.CadObject.Style = style;
 			}
 
-			if (builder.TryGetCadObject(this.DIMLDRBLK, out Block leaderArrow))
+			if (builder.TryGetCadObject(this.DIMLDRBLK, out BlockRecord leaderArrow))
 			{
 				this.CadObject.LeaderArrow = leaderArrow;
 			}
 
-			if (builder.TryGetCadObject(this.DIMBLK1, out Block dimArrow1))
+			if (builder.TryGetCadObject(this.DIMBLK1, out BlockRecord dimArrow1))
 			{
 				this.CadObject.DimArrow1 = dimArrow1;
 			}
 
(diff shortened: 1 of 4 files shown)


ACadSharp/IO/Templates/CadDimensionStyleTemplate.cs
```diff
@@ -32,17 +32,17 @@ public override void Build(CadDocumentBuilder builder)
 				this.CadObject.Style = style;
 			}
 
-			if (builder.TryGetCadObject(this.DIMLDRBLK, out Block leaderArrow))
+			if (builder.TryGetCadObject(this.DIMLDRBLK, out BlockRecord leaderArrow))
```
DomCR, line 35: For this kind of operations I've implemented a method to get a table entry and simplify the logging in case is not found, check `CadTemplate.getTableReference` you can use it to retrieve the `BlockRecord` in case of a not found it will notify to the user.

DomCR:
Great job with the documentation!!

mme1950:
Do you intent to replace all calls of TryGetCadObject with call getTableReference also when there is no name?

mme1950:
> Great job with the documentation!!

I hoped that would make it easier to use the propertries of DimenisonStyle for dimensions and leaders. But I still do not understand how to position dimension text properly.

DomCR:
> Do you intent to replace all calls of TryGetCadObject with call getTableReference also when there is no name?

`getTableReference` method is to wrap all the referencing to the tables and simplify the operations so is always logged when not found or there is some issue, in essence it does what all the same as TryGetCadObject but in one single place which it simplifies the implementation and the use.

I've decided to include the name and handle at the same time because dxf usually uses the name and dwg the handle but is not an strict rule so it checks for null values.

mme1950:
> I've decided to include the name and handle at the same time because dxf usually uses the name and dwg the handle but is not an strict rule so it checks for null values.

Then we also should change this line in `CadDimensionStyleTemplate.cs`:
`if (builder.TryGetCadObject(this.TextStyleHandle, out TextStyle style))`

DomCR (approved):

Issue #230, opened 2023-12-16
Leader: Style-Handle not resolved, AssociatedAnnotation missing


mme1950:
In DwgObjectReader.readLeader the handles of the associated `DimensionStyle` object (`template.DIMSTYLEHandle`) and the `AssociatedAnnotation` (`template.AnnotationHandle`) are read. But the `Build` method in `CadLeaderTemplate` is missing. Thus the references to the respective objects are not resolved.

An `AssociatedAnnotation` property must be added to the `Leader` class.

PR #243, opened 2024-01-06
#230 fixed

mme1950:
-  Leader: new property AssociatedAnnotation
-  CadLeaderTemplate: implementation added: Build

### DIFF
--- a/ACadSharp/Entities/Leader.cs
+++ b/ACadSharp/Entities/Leader.cs
@@ -106,7 +106,12 @@ namespace ACadSharp.Entities
 
 		//77	Color to use if leader's DIMCLRD = BYBLOCK
 
-		//340	Hard reference to associated annotation(mtext, tolerance, or insert entity)
+		/// <summary>
+		/// Hard reference to associated annotation (mtext, tolerance, or insert entity)
+		/// </summary>
+		[DxfCodeValue(340)]
+		public Entity AssociatedAnnotation { get; internal set; }
+
 
 		/// <summary>
 		/// Normal vector
@@ -132,7 +137,9 @@ namespace ACadSharp.Entities
 		[DxfCodeValue(213, 223, 233)]
 		public XYZ AnnotationOffset { get; set; } = XYZ.Zero;
 
+
 		private DimensionStyle _style = DimensionStyle.Default;
+
 
 		internal override void AssignDocument(CadDocument doc)
 		{
--- a/ACadSharp/Entities/Leader.cs
+++ b/ACadSharp/Entities/Leader.cs
@@ -109,7 +109,7 @@ namespace ACadSharp.Entities
 		/// <summary>
 		/// Hard reference to associated annotation (mtext, tolerance, or insert entity)
 		/// </summary>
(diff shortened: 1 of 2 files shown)


ACadSharp/Entities/Leader.cs
```diff
@@ -106,7 +106,12 @@ public DimensionStyle Style
 
 		//77	Color to use if leader's DIMCLRD = BYBLOCK
 
-		//340	Hard reference to associated annotation(mtext, tolerance, or insert entity)
+		/// <summary>
+		/// Hard reference to associated annotation (mtext, tolerance, or insert entity)
+		/// </summary>
+		[DxfCodeValue(340)]
```
DomCR, line 112: if this is a handle reference, please add the `DxfReferenceType.Handle` as a parameter of the `DxfCodeValue` attribute, thanks.

ACadSharp/IO/Templates/CadLeaderTemplate.cs
```diff
@@ -7,7 +8,24 @@ internal class CadLeaderTemplate : CadEntityTemplate
...
+			base.Build(builder);
+
+			Leader leader = (Leader)this.CadObject;
+
+			if (this.getTableReference(builder, this.DIMSTYLEHandle, string.Empty, out DimensionStyle style)) {
+				leader.Style = style;
+			}
+			if (builder.TryGetCadObject(this.AnnotationHandle, out Entity annotation)) {
+				leader.AssociatedAnnotation = annotation;
+			}
```
DomCR, line 28: The annotation is always present? if it is, it would be good to have an else with a log using the notification method so the user knows why the annotation is missing when the reader finishes.

mme1950: you provided the `getTableReference` method for table references. Shoud we implement a similar method for Entity references?

DomCR: I'm not 100% sure but I think that some entity references may have a null value, so I'm not sure that this method needs to be implemented, that's why I'm asking if the AssociatedAnnotation can be null or not, because if it can, there is no need to notify anything.

mme1950: We checked it with AutoCAD:
It is possible to create a LEADER without annotation. Then the field does not appear in DXF.
In this case the `AssociatedAnnotation` property should be null.

mme1950:
We checked it with AutoCAD:
It is possible to create a LEADER without annotation. Then the field does not appear in DXF.
In this case the `AssociatedAnnotation` property should be null.

ACadSharp/IO/Templates/CadLeaderTemplate.cs
DomCR: Exactly, that's why I've asked if this parameter is optional, because if it is, not finding the value or assigning a null value is a valid resolution for the reference and should not be notified as an error or warning.

DomCR:
Perfect! merging this PR.

Good work!!

PR #291, opened 2024-02-28
Fixed bug in reading bulges for Hatch.BoundaryPath.Polyline from DWG file

### LINKED ISSUE #282: Hatch - boundaryPath - Edges - Polyline
Hi,

**Describe the bug**
I have a problem with hatching, in particular with polyline boundaries, the Bulge parameter is set for only one point and there is no way to build the correct boundary for hatching

**To Reproduce**
Create a polyline shape in the drawing with more than one arc, hatch it and read the hatch data 

**Expected behavior**
A clear and concise description of what you expected to happen if needed.

**Screenshots**
It was expected that, like polyline in Entity, each point would have its own Bulge parameter.

SilverWolf2k20:
# Description

In the existing algorithm, when BoundaryPath of Polyline type had convexities, they were overwritten. This is incorrect according to the DWG format specification.

image

Below I offer my solution to the problem.

 - Fixed #282

# Notes for reviewer
 - I've changed the way the bulges are stored. Now they are stored as a list.
 - I had to change the record loop from foreach to for.

### DIFF
--- a/ACadSharp/Entities/Hatch.BoundaryPath.Polyline.cs
+++ b/ACadSharp/Entities/Hatch.BoundaryPath.Polyline.cs
@@ -1,6 +1,7 @@
 ﻿using ACadSharp.Attributes;
 using CSMath;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace ACadSharp.Entities
 {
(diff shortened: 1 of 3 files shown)


DomCR:
Good job!! 

Thanks for the contribution!

DomCR:
# TODO 

Update dxf reader/writer for Hatch.BoundaryPath.Polyline
- https://github.com/DomCR/ACadSharp/tree/Dxf-Hatch-BoundaryPath-Polyline-update-bulge

Issue #292, opened 2024-02-29
Rework Multileader


mme1950:
We recently implemented classes and a reader to handle `MultiLeader` entities in ACadSharp.

Using the MultiLeader data we learned a lot about the confusing data structures and values. So we try do document the properties to support usage.

And we found some issues, in particular bad or inconsistent naming:

- [x] #273
- [x] Misspelled enum identifier in `MultiLeaderPropertyOverrideFlags`
- [x] `MultiLeader`/`MultiLeaderAnnotContext`: `TextAlignment` property etc.
- [x]  `MultiLeader`/`MultiLeaderAnnotContext`: `AlignmentType`, `BlockContentConnection` property.
- [x] `MultiLeaderAnnotContext`: bad property visibility. no setters for `IList` properties
- [x] `MultiLeaderAnnotContext`: Property naming
- [x] `MultiLeaderAnnotContext`: Should `Location`, `Rotation` doubled to `TextLocation`, `TextRotation` and `ContentBlockLocation`, `ContentBlockRotation`?

Note: This issue infers a breaking change.

 Branch created: https://github.com/nanoLogika/ACadSharp/tree/20240225_mme_MultiLeader_rework

mme1950:
### TextAlignment etc.

The standard value for the `TextAlignment` property is defined in the `MultiLeaderStyle` object in the field with the DXF group code 176. This field can have the value 0. So it can be assumed that the enum values are 0=Left, 1=centered, 2=right.

The `MultiLeader` entity has a field with the DXF group code 175 which is unknown according to the OD_S_DWG documentation, but documented as _Text Alignment Type_ in the DXF reference AC1024 and newer. This field can have the values 0, 1, 2. So it can be assumed that the enum values are 0=Left, 1=centered, 2=right.

A field with the same meaning is found in the `MultiLeaderAnnotationContext` with the DXF group code 176 documented as _Text Align Type_ with the enum values 0=Left, 1=centered, 2=right in the OD_S_DWG documentation.

Both should be named as `TextAlignment`.

The multileader entity has a field with the DXF group code 179 documented as _Justification_ in the OD_S_DWG with the values 1=Left, 2=centered, 3=right, documented as _Text Attachment Point_ in the DXF reference.

[...]

Issue #273, opened 2024-02-06
MultiLeader.EnableDogleg always true


mme1950:
The `MultiLeader.EnableDogleg` property indicates whether the multilieader is to be drawn with or without a dogleg aka landing distance.

The two bits that represent the boolean values `EnableLanding` (DXF 290) and `EnableDogleg` (DXF 291) should follow `LeaderLineWeight` (DXF 171) and should be followed by `LandingDistance` (DXF 41).

Between `LeaderLineWeight` and `LandingDistance` 16 additional undocumented bits are found.

A call of `_objectReader.Advance(2)` after reading `EnableLanding` and `EnableDogleg` synchronizes after this "gap" so that `LandingDistance` is read correctly.

Obviously `_objectReader.Advance(2)` must be called before reading `EnableLanding` and `EnableDogleg` to obtain correct values of `EnableLanding` and `EnableDogleg`.

We will fix this issue and open a PR.

mme1950:
Fixed by #293.

PR #293, opened 2024-02-29
20240225 mme multi leader rework

mme1950:
Fixes issues #292 and #273.


mme1950:
Hi @DomCR,

I created detailed comments in `MultiLeader`and `MultiLeaderAnnotContext`.
But I would recommend to merge the current state and leave the comments for another PR.

DomCR:
> Hi @DomCR, thank you for resolving 16 spurious bits in readMultiLeader. You changed the restriction of MultiLeader to AutoCAD version 2010+. I am not sure it will work, since we found some inconsitancies in the documentation and we could not test it.

I've added tests to check the different versions, and I haven't modified the current flow besides the fix for the 16 bits, if you spot an issue with the new implementation report it as an issue and in the mean time I'll roll back the changes.

ACadSharp/Objects/MultiLeaderStyle.cs
```diff
@@ -26,223 +26,493 @@ public class MultiLeaderStyle : CadObject, IDictionaryEntry
...
 		/// </summary>
+		/// <remarks>
+		/// <para>
+		/// This value can be overridden by the <see cref="MultiLeaderAnnotContext.BlockContentScale"/> property
+		/// when the <see cref="MultiLeaderPropertyOverrideFlags.BlockContentScale"/> flag is set in the
+		/// <see cref="MultiLeader.PropertyOverrideFlags"/> property.
+		/// </para>
+		/// </remarks>
 		[DxfCodeValue(47, 49, 140)]
 		public XYZ BlockContentScale { get; set; }	//TODO: Change to 3 doubles values to better support the Dxf reading
```
DomCR, line 462: Can you apply this change in this PR?
> Change to 3 doubles values to better support the Dxf reading

ACadSharp/Objects/MultiLeaderAnnotContext.cs
```diff
@@ -260,18 +263,25 @@ public partial class MultiLeaderAnnotContext : CadObject
...
+		/// <summary>
+		/// Location
+		/// </summary>
+		[DxfCodeValue(15, 25, 35)]
+		public XYZ BlockContentLocation { get; set; }
 		/// <summary>
 		/// Scale vector
 		/// </summary>
 		[DxfCodeValue(16)]
 		public XYZ BlockContentScale { get; set; }
```
DomCR, line 278: Are the DXF values 16, 26, 36? if that's the case please add them.

ACadSharp/Objects/MultiLeaderStyle.cs
mme1950: OK

ACadSharp/Objects/MultiLeaderAnnotContext.cs
```diff
@@ -132,7 +133,7 @@ public partial class MultiLeaderAnnotContext : CadObject
 		/// Rotation (radians)
 		/// </summary>
 		[DxfCodeValue(42)]
-		public double Rotation { get; set; }
+		public double TextRotation { get; set; }
```
DomCR, line 136: If this is an angle, we should add the `DxfReferenceType.IsAngle` flag in the attribute.

DomCR:
If you make the changes I think this is ready to go!

Great job!!!

ACadSharp/Objects/MultiLeaderAnnotContext.cs
mme1950: Group code is 16 in Open Design Spec DWG and in the DXF reference. But it is clearly a 3BD named Scale Vector.
DXF files contain 16, 26, 36. So this is what is schould be.

BTW: This property of `MultiLeaderAnnotationContext` correponds to the `MultiLeaderStyle` property `BlockContentScale` with the group codes 47, 49, 140, see comment.
_Should we really slpit the `MultiLeaderStyle.BlockContentScale` property and `MultiLeaderAnnotationContext.BlockContentScale` as well?_ (see above).

ACadSharp/Objects/MultiLeaderAnnotContext.cs
mme1950: Done: 
* `MultiLeaderAnnotContext.TextRotation`
* `MultiLeaderAnnotContext.BlockContentRotation`
* `MultiLeader.BlockContentRotation`
* `MultiLeaderStyle.BlockContentRotation`

ACadSharp/Objects/MultiLeaderAnnotContext.cs
DomCR: knowing the DXF documentation this codes 47, 49, 140 do not belong to an XYZ component, so I would change them to make it easier for the DXF reader, or if they are directly corelated to the `MultiLeaderStyle` assigned, then they should be read-only for the context and return the right value.

That would be my approach, but this will not be relevant until the dxf reader/writer are in place, I don't mind to keep the current implementation with the note, I would add another note to know the relationship with the Style.

mme1950: > knowing the DXF documentation this codes 47, 49, 140 do not belong to an XYZ component

right.

This property of `MultiLeaderStyle` ist the standard scale vector for block content. `MultiLeaderAnnotationContext` has the same property , that overrides the standard values defined in `MultiLeaderStyle` when the respective `MultiLeaderPropertyOverrideFlag` is set. This property has the dxf-codes 16, 26, 36, that definitely indicate an XYZ-type.

So I would prefer to keep both properties as XYZ.
Maybe we should just now keep it as ist is and discuss it later.

mme1950:
Hi @DomCR,
could you please merge this branch. Then we can proceed to the writers.

DomCR (approved):

Issue #297, opened 2024-03-05
Implement DWG Writer for MultiLeader etc.


mme1950:
The DWG writer methods for MultiLeader, MultiLeaderStyle etc. are not yet implemented.

We will implement it and open a PR.

mme1950:
Hi @DomCR,

it seems that ALL "UNLISTED" entities do not appear in the documents _cadObjects.

DomCR:
Here is the problem: 
https://github.com/DomCR/ACadSharp/blob/1d3203b2c38bba9a2340e6258d5e8a1d78c1389a/ACadSharp/IO/DWG/DwgStreamWriters/DwgObjectWriter.Common.cs#L109

By now all UNLISTED are ignored because they have to be related to a `DxfClass` and is currently not impmlemented.

Here is how the `DwgReader` handles this objects:

https://github.com/DomCR/ACadSharp/blob/1d3203b2c38bba9a2340e6258d5e8a1d78c1389a/ACadSharp/IO/DWG/DwgStreamReaders/DwgObjectReader.cs#L953

A provisional solution could be to check the type of the `CadObject` in `writeCommonData` and if is a `MultiLeader` or `MultiLeaderStyle` find the related class in the `CadDocument` and write the type using the classNumber instead of the `ObjectType`. That way you will be able to bypass the switch and avoid any breaking changes for any other UNLISTED object that the file may contain.

I'll open a branch to take a look for a generic implementation.

mme1950:
Thanks, it works.

mme1950:
Hi @DomCR ,
we are trying to understand the concepts of writing DXF.

* What is the DxfClassMap exactly required for.
* Is it possible to use a generic DXF-writer method based on reflection and the `DxfCodeValue`-attributes?
* Oviously there is a reason to implement dedicated writer methods. 
* An XYZ can be written using the first group code, e. g. using +0 for x, +10 for y, +20 for z, right?
* It does not work for an XYZ having the group codes 48, 49, 140 as assigned to  `MultiLeaderStyle.BlockContentScale`, but it is possible to write x, y, z separately.
* Handles have to be written with `writeHandle`.
* It would be convenient to have a method to write `Color `properties.

mme1950:
Hi @DomCR,

thanks for merging #292, we will continue with the writers.
#311 should enable writing unlisted types without work-around, right?

DomCR:
> What is the DxfClassMap exactly required for?

[...]

PR #300, opened 2024-03-06
20240306 mme write multi leader

mme1950:
see: #297


mme1950:
We will re-create this PR when #293 is merged.

PR #323, opened 2024-03-22
20240322 mme write multi leader

mme1950:
Implemetation of DWG and DXF writers for MultLeader, MultiLeaderStyle
see #297


ACadSharp/Entities/MultiLeader.cs
```diff
@@ -345,7 +346,7 @@ public override CadObject Clone()
 
 			clone.ContextData = (MultiLeaderAnnotContext)this.ContextData?.Clone();
 
-			foreach (var att in BlockAttributes)
+			foreach (var att in BlockAttributes.ToList())
```
DomCR, line 349: Is there a particular case to add the `ToList()`?

ACadSharp/IO/DXF/DxfStreamWriter/DxfStreamWriterBase.cs
```diff
@@ -29,13 +31,28 @@ public void Write(int code, CSMath.IVector value, DxfClassMap map)
 			}
 		}
 
-		public void WriteHandle(int code, IHandledCadObject value, DxfClassMap map)
+		public void WriteCmColor(int code, Color color, DxfClassMap map = null)
```
DomCR, line 34: I need to check this but I think that the codification for a true color in a DXF file is not the same as in DWG.

ACadSharp/IO/DXF/DxfStreamWriter/DxfStreamWriterBase.cs
```diff
@@ -29,13 +31,28 @@ public void Write(int code, CSMath.IVector value, DxfClassMap map)
...
+			if (color.IsTrueColor) {
+				arr[0] = (byte)color.B;
+				arr[1] = (byte)color.G;
+				arr[2] = (byte)color.R;
+				arr[3] = 0b1100_0000;
 			}
-			else
+			else {
+				arr[3] = 0b1100_0001;
+				arr[0] = (byte)color.Index;
```
DomCR, line 46: This encoding is not valid in DXF, if you check a the base sample you will see that the layer '0' has the color White which is listed in the index 7. 

This code gives a different number instead.

ACadSharp/IO/DXF/DxfStreamWriter/IDxfStreamWriter.cs
```diff
@@ -15,6 +15,8 @@ internal interface IDxfStreamWriter : IDisposable
 		
 		void Write(int code, IVector value, DxfClassMap map = null);
 
+		void WriteCmColor(int code, Color color, DxfClassMap map = null);
```
DomCR, line 18: To keep the scope for this task I would suggest to remove this method, if is needed we can implement it in another PR with a dedicated test.

ACadSharp/Objects/MultiLeaderAnnotContext.cs
```diff
@@ -360,7 +361,7 @@ public override CadObject Clone()
 		{
 			MultiLeaderAnnotContext clone = (MultiLeaderAnnotContext)base.Clone();
 
-			foreach (var leaderRoot in LeaderRoots)
+			foreach (var leaderRoot in LeaderRoots.ToList())
```
DomCR, line 364: Is there a particular case to add the ToList()?

ACadSharp/IO/DXF/DxfStreamWriter/DxfSectionWriterBase.Entities.cs
```diff
@@ -16,13 +18,19 @@ protected void writeEntity<T>(T entity)
...
-				case MultiLeader:
 				case Wipeout:
 					this.notify($"Entity type not implemented : {entity.GetType().FullName}", NotificationType.NotImplemented);
 					return;
 			}
 
-			this._writer.Write(DxfCode.Start, entity.ObjectName);
+			if (entity.ObjectName == "MLEADER")
+			{
+				this._writer.Write(DxfCode.Start, "MULTILEADER");
```
DomCR, line 28: The `ObjectName` is the equivalent of the DxfName for that object, I've noticed that MLEADER is not used either in a DXF file or a `DxfClass`, it would simplify things if we change the token in `DxfFileToken.EntityMLeader` for the valid name which is MULTILEADER not the one found in the documentation website.

This change will also allow you to remove the mapping dictionary in the DwgObjectWriter _classNameMap.

ACadSharp/IO/DWG/DwgStreamWriters/DwgObjectWriter.Common.cs
```diff
@@ -115,7 +121,8 @@ private void writeCommonData(CadObject cadObject)
 			switch (cadObject.ObjectType)
 			{
 				case ObjectType.UNLISTED:
-					if (this._document.Classes.TryGetByName(cadObject.ObjectName, out DxfClass dxfClass))
+					string className = mapClassName(cadObject.ObjectName);
```
DomCR, line 124: Related:
https://github.com/DomCR/ACadSharp/pull/323/files#r1546788565

ACadSharp/Entities/MultiLeader.cs
mme1950: @nka1994, why did you add `ToList()`?

ACadSharp/Objects/MultiLeaderAnnotContext.cs
mme1950: see above.

ACadSharp/IO/DXF/DxfStreamWriter/DxfStreamWriterBase.cs
mme1950: I will check this. I tested it with several index and true colors and received the same values on writing as read.

mme1950: Hi @DomCR,
I thought I understand most of it but oviously I don't.

Here some examples (Created: DWG and DXF from AutoCAD, DXF saved with our `writeMultiLeader` method.

		Original					Written
	90 BL	-1033052384 = C26C DF20				-1033052384
	91 BL	-1073741824 = C000 0000				-1040187392 = C200 0000
  
	92 BL	-1033052384 = C26C DF20 -> True Color		-1033052384
	93 BL	-1056964608 = C100 0000 -> Index 0 = ByBlock	-1056964608

Written color values are essentiall ok, except the 91.
bit 0 of the byte 3 indicates, if set, that ist is an index color. "C" seems always to be there. What does bit 1 mean?

mme1950: [20240322_ml_with_style.zip](https://github.com/DomCR/ACadSharp/files/14835022/20240322_ml_with_style.zip)

ACadSharp/IO/DXF/DxfStreamWriter/IDxfStreamWriter.cs
mme1950: OK, I open an issue and PR.

ACadSharp/IO/DXF/DxfStreamWriter/DxfSectionWriterBase.Entities.cs
mme1950: done

ACadSharp/IO/DWG/DwgStreamWriters/DwgObjectWriter.Common.cs
mme1950: done.

ACadSharp/IO/DXF/DxfStreamWriter/DxfSectionWriterBase.Entities.cs
mme1950: done

ACadSharp/IO/DXF/DxfStreamWriter/DxfStreamWriterBase.cs
mme1950: Here we have again this case: 0xC0000000 seems to be something special. It looks like the true color black. But bit 1 is not set. Remember you hat this value for CECOLOR.

ACadSharp/IO/DXF/DxfStreamWriter/DxfSectionWriterBase.Entities.cs
```diff
@@ -649,6 +653,162 @@ private void writeMTextValue(string text)
 			this._writer.Write(1, text);
 		}
 
+		private void writeMultiLeader(MultiLeader multiLeader)
+		{
+			MultiLeaderAnnotContext contextData = multiLeader.ContextData;
+
+			this._writer.Write(160, 644);
```
DomCR, line 660: 160 is the code for the length of the preview bytes? if that's the case, is the same as for 310, should not be here.

ACadSharp/IO/DXF/DxfStreamWriter/DxfSectionWriterBase.Entities.cs
```diff
@@ -649,6 +653,162 @@ private void writeMTextValue(string text)
...
+			this._writer.Write(273, (short)contextData.TextTopAttachment);
+			this._writer.Write(301, "}");       //	CONTEXT_DATA
+		}
+
+		private void writeLeaderRoot(MultiLeaderAnnotContext.LeaderRoot leaderRoot)
+		{
+			this._writer.Write(302, "LEADER{");
+
+			this._writer.Write(290, true); // Has Set Last Leader Line Point
+			this._writer.Write(291, true); // Has Set Dogleg Vector
```
DomCR, line 784: Boolean values are written like:

```C#
value ? (short)1 : (short)0
```

ACadSharp/IO/DWG/DwgStreamReaders/DwgObjectReader.cs
```diff
@@ -2935,7 +2935,7 @@ private CadTemplate readMultiLeader()
 			}
 
 			//	BL Number of Block Labels 
-			int blockLabelCount = this._objectReader.ReadBitShort();
+			int blockLabelCount = this._objectReader.ReadBitLong();
```
DomCR, line 2938: I guess that this line has been always been wrong but the value was always 0 so that's why we haven't had any problem, can you confirm it?

ACadSharp/IO/DWG/DwgStreamWriters/DwgObjectWriter.Objects.cs
```diff
@@ -25,9 +25,7 @@ private void writeObject(CadObject obj)
 			switch (obj)
 			{
 				case Material:
-				case MultiLeaderStyle:
 				case MultiLeaderAnnotContext:
-				case SortEntitiesTable:
```
DomCR, line 30: Please re-add the `SortEntitiesTable` to be ignored, I still have to take a look at this one, thanks.

ACadSharp/IO/DXF/DxfStreamWriter/DxfSectionWriterBase.Entities.cs
```diff
@@ -14,7 +16,7 @@ protected void writeEntity<T>(T entity)
 			switch (entity)
 			{
 				case Solid3D:
-				case MultiLeader:
+				case Wipeout:
```
DomCR, line 19: Remove this line please, `Wipeout` should not be ignored.

ACadSharp/IO/DXF/DxfStreamWriter/DxfSectionWriterBase.cs
```diff
@@ -67,6 +57,24 @@ protected void writeCommonObjectData(CadObject cadObject)
 				//Add the dictionary in the object holder
 				this.Holder.Objects.Enqueue(cadObject.XDictionary);
 			}
+
+			if (cadObject.Reactors != null && cadObject.Reactors.Count > 0) {
```
DomCR, line 61: I do not consider Reactors to be fully implemented, if is needed for your implementation leave the writer as it is, but just for you to know that they may change the shape/behavior in the future.

ACadSharp/Objects/MultiLeaderAnnotContext.cs
DomCR, line 1: If this objects lives in a `CadDictionary` it should inherit the class `NonGraphicalObject`, check the PR as an example:
https://github.com/DomCR/ACadSharp/pull/374
If the class is not inherit the dictionary won't be able to add the entry.

DomCR (approved):

DomCR:
Approved and merged! 

Good work!

PR #355, opened 2024-06-03
fixed xrecord data reading

Martmath:
# Description

Fix the XRecord object reading for dwg files.


DomCR:
Hi @Martmath,

It seems that the `DwgReader` test have failed for your PR.

Could you also provide a description about what you want to fix/achieve, thanks.

Martmath:
I got an offset variable equal to the length of the data(~100) and the cursor this._objectReader.Position was about a few thousand in readrecord(). Tested on teigha DWG 2018 (nanocad 2022).  So, the condition
this._objectReader.Position < offset was always incorrect. Perhaps this is a feature of teigha dwg. Unfortunately, the error was at work, on a work file. I need to deploy the environment(vs /.net/nanocad to get the example file)at home for testing. So, I'm sorry, but I'll only have time to check this weekend or next.

DomCR:
You where right, the XRecord information was skipped all the time, I've pushed a fix for the missing implementations for the XRecord reading.

I couldn't find any information about the codes that you added 5001 and 5005, do you have any source of information to confirm these 2 codes?

Let me know if my implementation works for you.

Thanks for your contribution great job!

Martmath:
As I understand, Autodesk recommends 500x codes for storing ARX object data. I took them from [https://help.autodesk.com](https://help.autodesk.com/view/OARX/2022/ENU/?guid=GUID-9623358B-39B8-4A74-AAE4-038C42228A03). What I know, teigha apps use they for the same data types in a resbuffer too.  Since they are taken from the official website, I do not think that this part of code will create problems...but the problems as here will be solved: https://github.com/haplokuon/netDxf/issues/127

Martmath:
Also thank you for your great work! Life hack: if someone wants to read their ARX objects - now they can be serialized (and deserialized through your project) to xrecords))

DomCR:
Thanks for the tip @Martmath! I'll take a look at it!

PR #385, opened 2024-07-22
mleaderstyle defaults

DomCR:
# Description

Fix for MultiLeaderStyle DwgWriter.


mme1950:
MultiLeaderStyle:
Why are the properties BlockContent and ArrowHead of type BlockRecord not cloned?

PR #483, opened 2024-11-04
made the project compatible with netstandard2.0

mdmozibur:
# Description

just 4 file was affected and the changes are very small.


src/ACadSharp/ACadSharp.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
 	<PropertyGroup>
-		<TargetFrameworks>net5.0;net6.0;net7.0;net48;netstandard2.1</TargetFrameworks>
+		<TargetFrameworks>net5.0;net6.0;net7.0;net48;netstandard2.0</TargetFrameworks>
```
DomCR, line 4: I cannot see the reason of this downgrade.

Could you justify why should be implemented for 2.0 instead of 2.1?
Can you add 2.0 without removing the 2.1?

src/ACadSharp/IO/DWG/DwgStreamWriters/DwgHandleWriter.cs
```diff
@@ -30,7 +30,7 @@ public DwgHandleWriter(ACadVersion version, MemoryStream stream, Dictionary<ulon
 				this._handleMap.Add(item.Key, item.Value);
 			}
 #else
-			this._handleMap = new Dictionary<ulong, long>(map.OrderBy(o => o.Key));
```
DomCR, line 33: I think this fixed some issues for the dwg output, it cannot be removed.

src/ACadSharp/Objects/CadDictionary.cs
```diff
@@ -283,7 +283,14 @@ public bool ContainsKey(string key)
 		/// <returns>true if the element is successfully removed; otherwise, false.</returns>
 		public bool Remove(string key, out NonGraphicalObject item)
 		{
-			if (this._entries.Remove(key, out item))
```
DomCR, line 286: Net framework does not support this method either but there is an implementation in the [CSUtilities](https://github.com/DomCR/CSUtilities).

src/ACadSharp/IO/DWG/DwgStreamWriters/DwgHandleWriter.cs
mdmozibur: Can I get any example dwg file that faces issues with this being not changed?
I am doubting that because, you are creating a new Dictionary. The order of the collection that you are providing shouldn't have any effect because dictionaries are unordered collections.
If you can give me a dwg file with the issues mentioned, then I can dig deep into it.

DomCR:
Fix compatibility for netstandard20 without compromising the project or code.

- https://github.com/DomCR/ACadSharp/pull/506

PR #511, opened 2024-12-16
Annotative dimensions

DomCR:
# Description

Fix `DwgWriter` for annotative dimensions.


DomCR:
will be fixed in:

- https://github.com/DomCR/ACadSharp/pull/559

Issue #480, opened 2024-10-24
Create vertices from Spline


DomCR:
### Discussed in https://github.com/DomCR/ACadSharp/discussions/478

<div type='discussions-op-text'>

<sup>Originally posted by **JorenPelfrene** October 23, 2024</sup>
Hi, I find the PolyVertexes method very helpful for Arc and Ellipse entities. It would be great to also have this available for Spline (in the future).
It's a handy feature when using ACadSharp together with the Clipper2 library for performing boolean operations on 2D geometry.

Thank you for your great work ;)</div>

PR #586, opened 2025-03-20
issue-480_spline-polygonalvertices

JorenPelfrene:
# Description

Implement PolygonalVertexes method for Bsplines and adjust BoundingBox method

# Tasks done in this PR
* PolygonalVertexes method
* PointOnSpline method (public; could be useful)
* BoundingBox adjustment: no longer based on control points, but on discretization of curve (256 points)

- #480 
- #478


src/ACadSharp/Entities/Spline.cs
```diff
@@ -112,7 +114,82 @@ public Spline() : base() { }
 		/// <inheritdoc/>
 		public override BoundingBox GetBoundingBox()
 		{
-			return BoundingBox.FromPoints(this.ControlPoints);
+			List<XYZ> vertices = this.PolygonalVertexes(256);
+
+			return BoundingBox.FromPoints(vertices);
+
+			// previous implementation: control points can be outside the bounding box of the spline curve
+			//return BoundingBox.FromPoints(this.ControlPoints);
```
DomCR, line 122: Thanks for the fix, the comments can be removed.

src/ACadSharp/Entities/Spline.cs
```diff
@@ -112,7 +114,82 @@ public Spline() : base() { }
...
+		}
+
+		/// <summary>
+		/// Converts the spline in a list of vertexes.
+		/// </summary>
+		/// <param name="precision">Number of vertexes generated.</param>
+		/// <returns>A list vertexes that represents the spline expressed in object coordinate system.</returns>
+		public List<XYZ> PolygonalVertexes(int precision)
```
DomCR, line 130: netDxf has an equivalent method in the archive:
https://github.com/haplokuon/netDxf/blob/3d24d7c23418bd3a077d0e32a639f2ffe95bff51/netDxf/Entities/Spline.cs#L535

Can we add a test for this method? just to be sure that we are checking the correct functionality.

src/ACadSharp/Entities/Spline.cs
```diff
@@ -112,7 +114,82 @@ public Spline() : base() { }
...
+
+				return ocsVertexes;
+		}
+
+		/// <summary>
+		/// Gets a point on the spline.
+		/// </summary>
+		/// <param name="t">Parametric position on spline, between 0 and 1 (not linearly related to curve length).</param>
+		/// <returns>Point coordinates for the given parametric position on the spline.</returns>
+		public XYZ PointOnSpline(double t)
```
DomCR, line 153: Same as `PolygonalVertexes()`, it would be good to have a test for this one too.

src/ACadSharp/Entities/Spline.cs
```diff
@@ -112,7 +114,82 @@ public Spline() : base() { }
...
+				P += this.N(i, this.Degree, u) * this.ControlPoints[i];
+			}
+			return P;
+		}
+
+		// Bspline basis function N_{i,k}(u)
+		// u = parameter, ranges from lowest knot value to highest knot value
+		// k is the spline degree
+		// i ranges from 0 to (m-1) with m is the number of control points
+		internal double N(int i, int k, double u)
```
DomCR, line 171: If this method will only be used in `Spline` it should be `private`.

DomCR:
Great work! Thanks for the contribution!

PR #602, opened 2025-04-06
add lines example file

thaiguer:
# Description

Just add a new file in the examples, creating some lines.


DomCR:
Hi @thaiguer,

Thanks for the PR expanding the examples, I think that is what people looks the most when they want to use the library.

FYI, I started creating dedicated documents for the library https://github.com/DomCR/ACadSharp/tree/master/docs/

I don't know how this is received by the community but if you have an idea of how to increase the visibility it would be really helpful.

Thanks for your support!

PR #621, opened 2025-04-20
20250402 nka multileader clone bugfix

mme1950:
# Description
Various fails with MultiLeader Clone - Write - Read

# Tasks done in this PR
## Rework, implementation added:
-  Clone
-  AssignDocument, UnassignDocument
-  updateTable, updateCollection
-  onTableRemove
-  Additional boolean field in MultiLeaderStyle
-  FlowDirection.ByStyle = 5
-  Defaults added.


mme1950:
The MLeaderStyles collection contains a UnknownNonGraphicalObject with name Standard instead of an MultiLeaderStyle object. Do we have to correct the sample Documents?

src/ACadSharp/Entities/HatchPattern.cs
```diff
@@ -116,7 +116,7 @@ public static IEnumerable<HatchPattern> LoadFrom(string path)
 		/// </summary>
 		/// <param name="filename"></param>
 		/// <param name="patterns"></param>
-		public static void SavePatterns(string filename, params IEnumerable<HatchPattern> patterns)
+		public static void SavePatterns(string filename, params HatchPattern[] patterns)
```
DomCR, line 119: Is there a reason for this change? does it change a logic that I'm missing?

mme1950:
Sorry, I got a compilation error and quickly change it without inveszigating it. It is not essential.

src/ACadSharp/Entities/MultiLeader.cs
```diff
@@ -33,7 +39,14 @@ public partial class MultiLeader : Entity
 		/// <see cref="MultiLeaderAnnotContext.LeaderLine.OverrideFlags"/> property.
 		/// </remarks>
 		[DxfCodeValue(342)]
```
DomCR, line 41: If this is pointing to a handle can we add the flag `DxfReferenceType.Handle` to have the reference in the map?

src/ACadSharp/Entities/MultiLeader.cs
```diff
@@ -136,7 +149,26 @@ public partial class MultiLeader : Entity
 		/// <see cref="MultiLeaderAnnotContext.LeaderLine.OverrideFlags"/> property.
 		/// </remarks>
 		[DxfCodeValue(341)]
```
DomCR, line 151: Same as https://github.com/DomCR/ACadSharp/pull/621/files#r2055705987

src/ACadSharp/Entities/MultiLeader.cs
```diff
@@ -322,7 +372,26 @@ public partial class MultiLeader : Entity
 		/// is assumed to be used.
 		/// </remarks>
 		[DxfCodeValue(343)]
-		public TextStyle TextStyle { get; set; }
+		public TextStyle TextStyle
```
DomCR, line 375: Same as https://github.com/DomCR/ACadSharp/pull/621/files#r2055705987

src/ACadSharp/Entities/MultiLeader.cs
```diff
@@ -73,7 +86,7 @@ public partial class MultiLeader : Entity
 		/// <summary>
 		/// Contains the multileader content (block/text) and the leaders.
 		/// </summary>
-		public MultiLeaderAnnotContext ContextData { get; set; }
+		public MultiLeaderAnnotContext ContextData { get; set; } = new MultiLeaderAnnotContext();
```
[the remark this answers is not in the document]
mme1950: No, it is an embedded data structure and surely not to be used or registered somewhere else.
I have seen additional "complete" `MultiLeaderAnnotContext` objects in DXF, I think with the same data.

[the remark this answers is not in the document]
DomCR: Taking a closer look, `MultiLeaderAnnotContext` is not a `NonGraphicalObject` it only contains information from the `MultiLeader`, but it doesn't have a handle, owner  or needs to be assigned to the document.

src/ACadSharp/Entities/MultiLeader.cs
```diff
@@ -517,15 +594,83 @@ public override CadObject Clone()
...
+		{
+			base.AssignDocument(doc);
+
+			this._style = this.updateCollection<MultiLeaderStyle>(this._style, doc.MLeaderStyles);
+			this._textStyle = this.updateTable(this._textStyle, doc.TextStyles);
+			this._leaderLineType = this.updateTable(this._leaderLineType, doc.LineTypes);
+			this._arrowhead = this.updateTable(this._arrowhead, doc.BlockRecords);
+			this._blockContent = this.updateTable(this._blockContent, doc.BlockRecords);
+
+			this.ContextData.AssignDocument(doc);
```
DomCR, line 622: `ContextData` is not a `CadObject`, so it doesn't need to be assigned into the document.

src/ACadSharp/Entities/MultiLeader.cs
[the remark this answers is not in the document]
DomCR: I wouldn't allow null values to ensure that the writers work properly.

DomCR:
Looks good, and actually completes the work from:

- https://github.com/DomCR/ACadSharp/pull/385

I'll review any changes you make and approve this one once is ready.

Great job!

src/ACadSharp/Entities/HatchPattern.cs
mme1950: Sorry, I change it to avoud a compilation error, without investigation. Not essential!

src/ACadSharp/Entities/MultiLeader.cs
mme1950: Yes, I change it.

src/ACadSharp/Entities/MultiLeader.cs
```diff
@@ -216,7 +248,25 @@ public partial class MultiLeader : Entity
 		/// for this <see cref="MultiLeader"/>.
 		/// </summary>
 		[DxfCodeValue(340)]
-		public MultiLeaderStyle Style { get; set; }
+		public MultiLeaderStyle Style {
```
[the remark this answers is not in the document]
mme1950: Done.

src/ACadSharp/Entities/MultiLeader.cs
mme1950: Done

src/ACadSharp/Entities/MultiLeader.cs
mme1950: We have to call `ContextData.AssignDocument` to handle referenced `BlockRecord`, `TextStyle`, and `LeaderRoot.LeaderLine.LineStyle`.

mme1950:
The ContextData object is a sub-structure of the MultiLeader. It is to be created when the MultiLeader is created and to be cloned when the MultiLeader is cloned. The public ContextData property should be read-only.
Then the reader must be changed. It must not create a new MultiLeaderAnnotContext but fill the ContextDate that was created by the MultiLeader object.

mme1950:
Hi @DomCR 
We support MultiLeaderStyle for R2010+ = AC1024+. Some dynamic-blocks tests with samples AC1021, AC1015, AC1014 fail.
What shall we do?

src/ACadSharp/Objects/MultiLeaderAnnotContext.cs
```diff
@@ -7,15 +8,17 @@
...
 
-namespace ACadSharp.Objects
-{
+namespace ACadSharp.Objects {
 
 	/// <summary>
 	/// This class represents a subset ob the properties of the MLeaderAnnotContext
 	/// object, that are embedded into the MultiLeader entity.
 	/// </summary>
 	public partial class MultiLeaderAnnotContext : NonGraphicalObject
```
DomCR, line 17: `MultiLeaderAnnotContext` should not inherit `NonGraphicalObject` because is not a `CadObject` with an assignable handle.

src/ACadSharp/Entities/MultiLeader.cs
DomCR: We should try to link the properties from the `ContextData` to the document but with the context not being a `CadObject`.

src/ACadSharp/Objects/MultiLeaderAnnotContext.cs
DomCR: Nevermind, leave the `NonGraphicalObject` and the AssignDocument() to make it simple, I'll think of a refactor after the merge, I don't want to get stuck on this particular issue.

DomCR:
> Hi @DomCR We support MultiLeaderStyle for R2010+ = AC1024+. Some dynamic-blocks tests with samples AC1021, AC1015, AC1014 fail. What shall we do?

Apply this change in `DwgObjectReader`:

```C#
case "MLEADERSTYLE":
	if (!this.R2010Plus)
	{
                this.notify($"MLEADERSTYLE is not supported for {this._version}.", NotificationType.Warning);
		return null;
	}
	template = this.readMultiLeaderStyle();
	break;
```

Is not an unknown object, the reason why is not read is because we lack documentation, so it makes sense to skip it.

Note: I've tested removing the `if (!this.R2010Plus)` from the `readMultiLeaderStyle()` and it has passed all the test, if you remove the version guard and passes the test along with some test files in your end we could add the support for the older versions.

mme1950:
I do not remember why we restricted the MultiLeaderStyle reader to 2010+.
The DXF Reference 2005 does not mention MultiLeader etc.
In the DXF Reference 2008 and 2010 MultiLeader and MultiLeaderStyle is described. But MLEADER has no Style group code. But this may be simply wrong.
In the DXF Reference 2011 and later the DXF group codes become similar to the current version.

So maybe we should change the restriction 2008+ i. e. AC1018 and remove the MultiLeader tests for older versions.

DomCR:
I would delete the flag by now, it passes all the tests and it doesn't seem to cause any issue, if there is an error during the reading it will trigger the failsafe and continue. 

Before merging this PR to master I'll test it against my local samples and see if it triggers something unexpected, if that's the case, we can think of a fix for it.

mme1950:
Hi @DomCR, locally all tests succeed, why do they fail here?

DomCR:
My bad, in one of the lasts PRs that I merged I restored a flag in the test that was wrongly set, please apply this fix for the style:

```C#
case "MLEADERSTYLE":
	if (!this.R2010Plus)
	{
                this.notify($"MLEADERSTYLE is not supported for {this._version}.", NotificationType.Warning);
		return null;
	}
	template = this.readMultiLeaderStyle();
	break;
```

Apologies for the mistake.

src/ACadSharp/IO/DWG/DwgStreamReaders/DwgAnalyseTool.cs
```diff
@@ -0,0 +1,128 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace ACadSharp.IO.DWG.DwgStreamReaders {
+	internal static class DwgAnalyseTool {
```
DomCR, line 9: Is there a difference between these exploration methods and `DwgStreamReaderBase.Explore`?

mme1950: I had a problem (NPE) with `DwgStreamReaderBase.Explore` so I continued with my DwgAnalyseTool, that is does not try to read text fields etc. I simply forgot to take it out.

mme1950: I will remove this class.

DomCR:
Looks good! can we close this one?

- https://github.com/DomCR/ACadSharp/pull/385

mme1950:
Yes, I think everything you intended is inkludiert in the other PR.

mme1950:
But we still do not understand why MultiLeaders that wäre sauer by ACadSharp are invisible in AutoCAD.

DomCR:
> But we still do not understand why MultiLeaders that wäre sauer by ACadSharp are invisible in AutoCAD.

Not sure about this one, I'll take a look and see what I can find, do you want to keep this PR open for the investigation or can we merge this and create a new one?

mme1950:
Please merge this one. Thank you very much for your effort on Saturday.
You already opened #632,

PR #641, opened 2025-04-27
MultiLeader SPIKE

DomCR:
# Description

Investigation for MLeader objects not being correctly read by the `DwgReader` and written by `DwgWriter`.

# Tasks done in this PR
* [ ] DwgReader fix for MLeaderStyle.
* [ ] DwgWriter fix for MLeaderStyle.
* [ ] MLeader must be visible.
* [ ] Check defaults and values.

# Notes for reviewer

- After reading the OpenDesign docs I've found some errors in the code; a missing flag version and some wrongly read values that where read in the wrong format.
- The flag set as **IF IsNewFormat OR DXF file** has been removed because it was set as R2010 instead of a value in the APPID “ACAD_MLEADERVER”. This change seems to fix the older versions of the dwg but the value still is unknown for the reader and in some cases may cause to trigger an exception in the reader (not a breaking change if the Failsafe is on).
- The last bit `UnknownFlag298` seems to be present only for R2013Plus versions of dwg, this has been found by a manual investigation which may be wrongly assumed, to avoid the wrong values it will be set as an internal value.

### DIFF
--- a/src/ACadSharp/IO/DWG/DwgStreamReaders/DwgObjectReader.cs
+++ b/src/ACadSharp/IO/DWG/DwgStreamReaders/DwgObjectReader.cs
@@ -1054,10 +1054,6 @@ namespace ACadSharp.IO.DWG
 					template = this.readMultiLeader();
 					break;
 				case "MLEADERSTYLE":
-					if (!this.R2010Plus) {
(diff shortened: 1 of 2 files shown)


mme1950:
My problems with multileaders are solved with https://github.com/DomCR/ACadSharp/pull/658.
However you added some things, such as property default values, that my be helpful anyhow.

PR #646, opened 2025-04-29
Escape special characters when writing strings to svg

carlokok:
Use writestring to allow escaping special symbols.

# Description

When a string/text contains & or other special symbols it should be escaped, not written raw to svg.

# Tasks done in this PR
* [ ] Something awesome.


DomCR:
Looks good! thanks for the contribution!

DomCR (approved):

Issue #632, opened 2025-04-25
MultiLeader Block update


DomCR:
**Is your feature request related to a problem? Please describe.**
MultiLeader are not shown when are allocated in blocks.

**Describe the solution you'd like**
Create the MultiLeader block in a similar manner than dimensions so they are updated when they are allocated inside blocks.

**Additional context**
- https://github.com/DomCR/ACadSharp/issues/629

mme1950:
Hi @DomCR,
actually the a MultiLeader dos not show up, also when it is not in a block.
The attached files contain the following steps
* MultiLeader created in AutoCAD
* Read and Saved by ACadSharp
* Repaired in AutoCAD (Style colors reset to ByLayer, Style changed and reset.)

[m-ACadSharp-AutoCAD.zip](https://github.com/user-attachments/files/19906852/m-ACadSharp-AutoCAD.zip)

mme1950:
When I compare documents stored as DXF I see some differences:
* Colour are changed: bit 02000000 is added.
   C0000000 --> C2000000
   C1000000 --> C3000000
   It seems we have learn what the 02000000 bit means.
   But I do not think that the changes colours are relevant.
* The color of the leader line (BL 92) was null but is set to C3000000
   This is wrong, but should not be relevant.

mme1950:
In the original DXF The extended data of the multiLeader contain the following record:
1001	S	AcDbBlockRepETag	
1070	BS	1	
1071	BL	298
1005	S	2F4

1005 2F4 is the Handle of the MultiLeaderItself.

In the data written by ACadSharp these data contain:
1001	S	AcDbBlockRepETag	
1070	BS	1	
1071	BL	298	12A
1005	S	F402000000000000

**The first two bytes are flipped.**

I do not know what this means, but this seems to be a relevant difference.

mme1950:
Hi @DomCR,
as far I remember, you did not have additional information about CmColor bit 0x02000000,
`ReadCmColor` does not take this bit into account.
In AutoCAD the color is ByLayer, but is written as 0xC0000000. Since bit 0x01000000 is not set, this is read as true color black.
This is consequently written as 0xC300000.
Normally I expect that ByLayer is to be represented as index color 256.
To preserve the value of 0xC0000000 this has to be handled as a special case.
I try to find out if it makes any difference.

mme1950:
Hi @DomCR,
[...]

Issue #629, opened 2025-04-24
Some entities in generated DWG are invisible


mme1950:
I am creating a new CAD document by the following mechanisms
* clone a BlocksRecord from a source document.
* create a new BlockRecord filled with cloned Entities from a dynamic-block state from a source document.
* create a MText entitiy and add it to the documents entity list.

Problems:
* Dimension entities in a cloned block are not visible.
  When I edit the block in AutoCAD I see the Dimension. When I change the style and then reselect the original style and close the block editor, the Dimension becomes visible in the document.
* Multileader entities in a cloned block are not visible.
  When I edit the block in AutoCAD the MultiLeader is invisible. I can select all (Ctrl-A) including the MultiLeader. When I change the style and reselect reselect the original style the MultiLeader becomes visible, after closing the clock editor also in the document.
* A MText created using ACadSharp directly added to document's entities ist invisible.
  I can select the MText, when I try to edit it (context menu) it becomes visible.

Did anyone have the same or similar problems and found a solution?

mme1950:
Investigation
* Create an MText in AutoCAD
* Clone this MText using ACadSharp -> The cloed MText is visible in AutoCAD
* Create an MText using ACadSharp, set DrawingDirection = LeftToRight and TextHeight -> The Created MText is invisible in AutoCAD
   Editing the MText in AutoCAD makes the MText visible.
   The MText property grid dos not show any difference.
* Comparing the Properties of the MText created by AutoCAD and the MText created by ACadSharp shows a difference of the property AlignmentPoint (0,0,0).
* Set AlignmentPoint to (1,0,0) -> MText becomes visible in AutoCAD.

@DomCR Should we initialize DrawingDirection and AlignmentPoint?

mme1950:
Investigation Dimension
* Dimensions created by AutoCAD added to the document's entities are visible in AutoCAD. The Dimension's Block property is null.
* Dimensions that ere members of a cloned BlockRecords are not visible in AutoCAD. The Dimension's Block property is null.
* Edit the inserted BlockRecord in AutoCAD: change style and restore previous style of the Dimension, Then save the blockRecord. The Dimension becomes visible.
* Comparing all Properties exposed by ACadSharp of the Dimension before and after it was edited shows a difference; The Dimension's Block property is set.

DomCR:
Hi @mme1950,

**Dimensions**: this is a known issue, here is more info about it:
- https://github.com/DomCR/ACadSharp/discussions/556
[...]

PR #658, opened 2025-05-07
Support ACDB_MLEADEROBJECTCONTEXTDATA_CLASS

mme1950:
-  MultiLeaderObjectContextData (formally MultiLeaderAnnotContext) as standalone object.

* https://github.com/nanoLogika/ACadSharp/issues/10
* #632
* #629

### DIFF
--- a/src/ACadSharp/Classes/DxfClassCollection.cs
+++ b/src/ACadSharp/Classes/DxfClassCollection.cs
@@ -426,6 +426,18 @@ namespace ACadSharp.Classes
 				ProxyFlags = ProxyFlags.EraseAllowed | ProxyFlags.DisablesProxyWarningDialog,
 				WasZombie = false,
 			});
+
+			//AcDbMLeaderObjectContextData
+			doc.Classes.AddOrUpdate(new DxfClass {
+				CppClassName = DxfSubclassMarker.MultiLeaderObjectContextData,
+				ClassNumber = (short)(500 + doc.Classes.Count),
+				DwgVersion = ACadVersion.MC0_0,
+				DxfName = DxfFileToken.ObjectMLeaderContextData,
+				ItemClassId = 499,
+				MaintenanceVersion = 0,
+				ProxyFlags = ProxyFlags.EraseAllowed | ProxyFlags.DisablesProxyWarningDialog,
+				WasZombie = false,
+			});
 		}
 
 		/// <summary>
--- a/src/ACadSharp/Entities/MultiLeader.cs
+++ b/src/ACadSharp/Entities/MultiLeader.cs
@@ -34,9 +34,9 @@ namespace ACadSharp.Entities
 		/// </summary>
 		/// <remarks>
 		/// The value for all leader lines can be overridden for each individual leader line by the
-		/// <see cref="MultiLeaderAnnotContext.LeaderLine.Arrowhead"/> property when the
+		/// <see cref="MultiLeaderObjectContextData.LeaderLine.Arrowhead"/> property when the
 		/// <see cref="LeaderLinePropertOverrideFlags.Arrowhead"/> flag is set in the
-		/// <see cref="MultiLeaderAnnotContext.LeaderLine.OverrideFlags"/> property.
+		/// <see cref="MultiLeaderObjectContextData.LeaderLine.OverrideFlags"/> property.
 		/// </remarks>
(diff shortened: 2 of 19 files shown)


mme1950:
see #707

mme1950:
Hi @DomCR,

I changed MutiLeaderStyle:
- member sorting
- property default values (should be helpful, when a MultiLeaderStyle is created.)

DomCR:
Hi @mme1950,

For me it looks good, I'll review the PR now.

src/ACadSharp/Entities/MultiLeader.cs
```diff
@@ -86,7 +87,16 @@ public BlockRecord Arrowhead
...
 		/// Contains the multileader content (block/text) and the leaders.
 		/// </summary>
-		public MultiLeaderAnnotContext ContextData { get; private set; } = new MultiLeaderAnnotContext();
+		public MultiLeaderObjectContextData ContextData {
+			get {
+				if (tryGetContextdata(out MultiLeaderObjectContextData contextData)) {
+					return contextData;
+				}
+				else {
+					return _contextData;
```
DomCR, line 96: This field doesn't seem to be assign anywhere except by default, I would suggest the `BlockRecord.EvaluationGraph` approach, where the property always gets the dictionary entry if exists.

src/ACadSharp/Entities/MultiLeader.cs
```diff
@@ -607,9 +617,34 @@ public override CadObject Clone()
 			clone._arrowhead = (BlockRecord)this._arrowhead?.Clone();
 			clone._blockContent = (BlockRecord)this._blockContent?.Clone();
 
+			if (tryGetContextdata(out MultiLeaderObjectContextData contextData)) {
```
DomCR, line 620: This if can be removed if the property `ContextData` follows the indicated implementation, the dictionary will be cloned anyway so all the entries will be in the clone.

mme1950: OK, I have test it in the context of my application.

mme1950:
Thank you for the review.

src/ACadSharp/Classes/DxfClassCollection.cs
```diff
@@ -428,6 +428,17 @@ public static void UpdateDxfClasses(CadDocument doc)
 				ProxyFlags = ProxyFlags.EraseAllowed | ProxyFlags.DisablesProxyWarningDialog,
 				WasZombie = false,
 			});
+            //AcDbPdfReference
```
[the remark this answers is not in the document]
DomCR: The AcDbPdfReference is added to the collection 2 times.

src/ACadSharp/Entities/MultiLeader.cs
DomCR: Then that's a problem 🥲  

I haven't seen this entry myself so I'm not sure how it works, but I think it may be an override for the MLeader properties.
If that's the case, we should reflect that in the code, which means that the field _contextData should not be the same as the one in the dictionary. So we should keep it as it is and maybe add a shortcut to the one in the Dictionary with another property or a method, but it would be best if we know exactly what that entry in the dictionary does to the entity.

src/ACadSharp/Classes/DxfClassCollection.cs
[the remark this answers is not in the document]
mme1950: Duplicate lines removed.
Comment corrected.

src/ACadSharp/Entities/MultiLeader.cs
mme1950: What do you think about this approach?

Step 1:
* The embedded context-data object and the dictionary entries shall be handled as independent data structures.
* The clone must support multiple dictionary entries.

Step 2:
* A change of `EnableAnnotationScale` to true creates a dictionary entry, a change to false clears the dictionary.
* Method `AddAnnotationScale` and `RemoveAnnotationScale` should be added to create and remove additional dictionary entries. The createion of the first dictionary and the removal of the last entry must set the value of `EnableAnnotationScale`.

Step 3:
* Synchronization of property values of `MultiLeader` and the context-data objects shall be implemented.
  This is a big task.

DomCR:
Done!

Sorry for the delay.

Issue #626, opened 2025-04-23
Dynamic blocks: Flip Action


mme1950:
A dynamic block with a flip action can be read when the at least the following types are supported:

* BlockFlipAction
* BlockFlipParameter
* and base classes

Data from the BlockFlipParameter object can be used apply a transform to the entities listed in the FlipAction object.

mme1950:
We have set up the classes and readers.
Next step is to implement the transform.
We will open a PR.

mme1950:
Hi @DomCR,
a flip action is defined by two points defining a mirror axis. What do you think is the best way to set up a Transform?
I implemented a simple flip action with a vertical mirror axis through (0,0,0): entity.ApplyScale(new XYZ(-1, 1, 0), XYZ.Zero).

Some Remarks:
* Spline.ApplyTransform fails with "Index was out of range ..." because the loop over fit points runs to this.FitPoints.Capacity.
* A left-attached MText is not changed to right-attached.
* The entities in the block of an Insert are not transformed.

mme1950:
Fixed with #665.

PR #665, opened 2025-05-09
20250509 mme #626 11 support dynamic block flip actions

mme1950:
Tasks done in this PR

- [x] Classes `BlockFlipAction`, `BlockFlipParameter` and superclasses
- [x] reader method, template classes

### DIFF
--- a/src/ACadSharp/IO/Templates/CadBlockVisibilityParameterTemplate.cs
+++ b/src/ACadSharp/IO/Templates/CadBlockVisibilityParameterTemplate.cs
@@ -3,6 +3,8 @@ using System.Linq;
 using ACadSharp.Entities;
 using ACadSharp.Objects.Evaluations;
 
+using static ACadSharp.IO.Templates.CadBlockVisibilityParameterTemplate;
+
 namespace ACadSharp.IO.Templates
 {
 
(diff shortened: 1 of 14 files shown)


src/ACadSharp/Objects/Evaluations/BlockBasePointParameter.cs
```diff
@@ -0,0 +1,44 @@
...
+	public class BlockBasePointParameter : Block1PtParameter {
+
+		/// <inheritdoc/>
+		public override string ObjectName => DxfFileToken.ObjectBlockBasePointParameter;
+
+		/// <inheritdoc/>
+		public override string SubclassMarker => DxfSubclassMarker.BlockBasePointParameter;
+
+
+		[DxfCodeValue(1011, 1021, 1031)]
```
DomCR, line 37: This codes are used for `ExtendedData` are we sure that they have an assigned property in the object?

src/ACadSharp/IO/Templates/CadBlockVisibilityParameterTemplate.cs
DomCR, line 1: I know that this file contains multiple classes but can we split it in the different classes to make it easier to read?

src/ACadSharp/Objects/Evaluations/BlockBasePointParameter.cs
mme1950: hmm ...
(Formatted) DXF looks like this:

0	S	BLOCKBASEPOINTPARAMETER
5	H	7F954
330	H	7F953
100	S	AcDbEvalExpr
90	BL	1
98	BL	33
99	BL	329
100	S	AcDbBlockElement
300	S	Basispunkt
98	BL	33
99	BL	329
1071	BL	0
100	S	AcDbBlockParameter
280	BS	1
281	BS	0
100	S	AcDbBlock1PtParameter
1010	BD	0.0
1020	BD	0.0
1030	BD	0.0
93	BL	0
170	BS	0
171	BS	0
100	S	AcDbBlockBasepointParameter
1011	BD	0.0
1021	BD	0.0
1031	BD	0.0
1012	BD	0.0
1022	BD	0.0
1032	BD	0.0

src/ACadSharp/IO/Templates/CadBlockVisibilityParameterTemplate.cs
mme1950: OK, done.

src/ACadSharp/Objects/Evaluations/BlockBasePointParameter.cs
mme1950: I made it similar to `Block1PtParameter` and `Block2PtParameter` (not implemented by me).
[1010, 1020, 1030] are the coordinates of THE point in `Block1PtParameter`.

mme1950:
Currently I am `BlockFlipAction` and `BlockFlipParameter` to evaluate a flipped state.
This can be used to convert a flipped state of the block to SVG.
I also use it to insert the flipped entities in a new BlockRecord into a new DWG.
This works fine. 
The `BlockBasepointParameter` is contained in the DXF but not needed for these tasks.

DomCR:
Looks good! Great work!
 
There are some missing DxfCodes missing but is ok if there is no support for dxf.

Issue #672, opened 2025-05-15
Dimension Anonymous Block:


mme1950:
Some features related to the anonymous Block of a `Dimension` seem to be missing or incomplete:

1. The BlockRecord referenced by the `Dimension.Block` property is not cloned.
2. The `Dimension.Block` is always null, when it is set, because the call of updateTable is wrong.
3. The anonymous BlockRecord must be exclusively related to one `Dimension`. When dimensions originating from more than one source are to be combined in a new document it must ensured that `BlockRecord` names are unique. `updateTable` cannot be used.
4. The reference to the Block is not witten:
```
//this._writer.HandleReference(DwgReferenceType.HardPointer, dimension.Block);
this._writer.HandleReference(DwgReferenceType.HardPointer, null);
```

DomCR:
Hi @mme1950,

This feature is being implemented in this PR:

- https://github.com/DomCR/ACadSharp/pull/559

I'll add the fix related to the block in another sub PR so is fixed beforehand.

PR #673, opened 2025-05-15
Proposed fixes for #672

mme1950:
see #672.

### DIFF
--- a/src/ACadSharp/Entities/Dimension.cs
+++ b/src/ACadSharp/Entities/Dimension.cs
@@ -1,5 +1,7 @@
 ﻿using ACadSharp.Attributes;
 using ACadSharp.Tables;
+using ACadSharp.Tables.Collections;
+
 using CSMath;
 using CSUtilities.Extensions;
 using System;
@@ -30,9 +32,9 @@ namespace ACadSharp.Entities
 		public BlockRecord Block
 		{
 			get { return this._block; }
-			set
-			{
-				this._block = this.updateTable(this._block, this.Document?.BlockRecords);
+			set {
+				this._block = value;
+				this.addToBlockRecordsTable(_block);
 			}
 		}
 
@@ -240,6 +242,7 @@ namespace ACadSharp.Entities
 			Dimension clone = (Dimension)base.Clone();
 
 			clone.Style = (DimensionStyle)(this.Style.Clone());
+			clone.Block = (BlockRecord)this.Block?.Clone();
 
 			return clone;
 		}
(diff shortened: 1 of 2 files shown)


src/ACadSharp/IO/DWG/DwgStreamWriters/DwgObjectWriter.Entities.cs
```diff
@@ -342,8 +342,8 @@ private void writeCommonDimensionData(Dimension dimension)
 			this._writer.HandleReference(DwgReferenceType.HardPointer, dimension.Style);
 			//H 2 anonymous BLOCK(hard pointer)
 			//TODO: fix annotative dimensions
-			//this._writer.HandleReference(DwgReferenceType.HardPointer, dimension.Block);
-			this._writer.HandleReference(DwgReferenceType.HardPointer, null);
+			this._writer.HandleReference(DwgReferenceType.HardPointer, dimension.Block);
+			//this._writer.HandleReference(DwgReferenceType.HardPointer, null);
```
DomCR, line 346: This line can be removed, no need to keep it as a comment.

src/ACadSharp/Entities/Dimension.cs
```diff
@@ -280,5 +283,35 @@ protected override void tableOnRemove(object sender, CollectionChangedEventArgs
 				this._block = null;
 			}
 		}
+
+
+		private void addToBlockRecordsTable(BlockRecord block)
```
DomCR, line 288: I don't think that's a good practice to make changes to an object that is not managed by the library itself, I would suggest to keep the old method `updateTable` and add a guard that check that the block has the flag anonymous set.

If the naming of the block is an issue for any software, we can add a method that specifically changes the block name to a unique one for the current dimension, or a method in the `CadDocument` that updates the name for all anonymous blocks.

src/ACadSharp/IO/DWG/DwgStreamWriters/DwgObjectWriter.Entities.cs
mme1950: done

src/ACadSharp/Entities/Dimension.cs
mme1950: `updateTable` tries to find a `BlockRecord` with the same name and returns the existing one so that the new one is to be discarded. We have to ensure that the new one is kept - with a new name.

DomCR: I agree with that, I think that the solution should affect all the anonymous blocks, I'm thinking on modifying the Block record table to check for anonymous blocks and change the name to a new name when they are added into the table, we should also inform the user of this by documenting the corresponding flag inside the `BlockRecord`. 

Giving that this blocks are used only internally and created by the software on the spot, it would make sense that this block are partially handled by the library also, this solution offers the option for the user to change the block to a specific one (if someone wants the option) to change the dimensions for something else, at least until the graphic software takes over.

DomCR: It would not make much sense, if a dimension has its own geometry each one needs a separated block. Tables and MLeaders have to be the same.

In any case if the block is already in the table that one can be referenced again by the user.

mme1950:
This will be obsolete when #559 is completed.

Issue #682, opened 2025-06-13
Some entity do not have bounding box


vmair:
**Some entity do not have bounding box**
This entity is an arrow。
![Image](https://github.com/user-attachments/assets/f5cecc4f-2392-4083-bd73-fe759c31946a)

vmair:
**The Solid class does not implement the relevant methods**
`public override BoundingBox GetBoundingBox()
		{
			return BoundingBox.Null;
		}
`
It might be possible to implement the method via the corner point attribute

PR #685, opened 2025-06-16
Proposed fixes for #682

vmair:
Implement the GetBoundingBox method of the Solid class.


src/ACadSharp/Entities/Solid.cs
```diff
@@ -74,7 +75,22 @@ public override void ApplyTransform(Transform transform)
 		/// <inheritdoc/>
 		public override BoundingBox GetBoundingBox()
```
DomCR, line 76: The implementation looks good, but I think it can be simplified using the method `BoundingBox.FromPoints()` and add all the points of the `Solid`.

Also, I think we can remove the try catch, giving the nature of the XYZ being an structure with default value of {0,0,0} I don't see an unmanaged exception popping up.

DomCR:
The PR looks good, thanks for the contribution!

src/ACadSharp/Entities/Solid.cs
vmair: Relevant code has been optimized

DomCR (approved):

PR #736, opened 2025-08-16
Header property for system variable CMLEADERSTYLE

mme1950:
VariablesDictionary initialized with CMLEADERSTYLE='Standard'.
CurrentMultiLeaderStyle, CurrentMultiLeaderStyleName as new header properties.
Synchronization CurrentMultiLeaderStyleName <-> VaraibleDictionary

### DIFF
--- a/src/ACadSharp/Header/CadHeader.cs
+++ b/src/ACadSharp/Header/CadHeader.cs
@@ -1221,7 +1221,7 @@ namespace ACadSharp.Header
 				}
 				else
 				{
-					DictionaryVariable variableDictionaryEntry = ensureVariableDictionaryEntryExists(this.Document);
+					DictionaryVariable variableDictionaryEntry = ensureVariableDictionaryEntryExists(this.Document, CadDictionary.CurrentMultiLeaderStyle, MultiLeaderStyle.DefaultName);
 					return variableDictionaryEntry.Value;
 				}
 			}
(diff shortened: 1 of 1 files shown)


src/ACadSharp/Objects/CadDictionary.cs
```diff
@@ -185,7 +190,14 @@ public static void CreateDefaultEntries(CadDictionary root)
 			root.TryAdd(new CadDictionary(AcadPlotSettings));
 			// { AcadPlotStyleName, new CadDictionaryWithDefault() },	//Add default entry "Normal"	PlaceHolder	??
 
-			root.TryAdd(new CadDictionary(VariableDictionary));
+			CadDictionary variableDictionary = root.ensureCadDictionaryExist(VariableDictionary);
+			root.TryAdd(variableDictionary);
+			DictionaryVariable cmLeaderStyleEntry = new DictionaryVariable()
```
DomCR, line 195: Is the dictionary variable needed to ensure the right behavior of the CMLEADERSTYLE header variable to work? 

There is no code in place to manage this kind of variables, if this works I'm OK to merge it but we will need to have a proper solution for this.

mme1950: I tried to find CMLEADERSTYLE in DXF. The dictionary variable was the only occurrence.
In a document created in ACadSharp the current `MultiLeaderStyle` set is `Standard`.
Changing the dictionary variable sets the current `MultiLeaderStyle`. I do not know another mechanism.
You think of a more general approach to get and set values of such dictionary variables. Are there other ones?

mme1950: > Is the dictionary variable needed to ensure the right behavior of the CMLEADERSTYLE header variable to work?
yes.

src/ACadSharp/Header/CadHeader.cs
```diff
@@ -1194,6 +1195,78 @@ public string CurrentDimensionStyleName
...
+					variableDictionaryEntry.Value = value;
+				}
+				else
+				{
+					this._currentMultiLeaderStyle = new MultiLeaderStyle(value);
+				}
+			}
+		}
+
+		private static DictionaryVariable ensureVariableDictionaryEntryExists(CadDocument doc) {
```
DomCR, line 1251: To make this method more generic and be able to use for the future variables in the dictionary we should parametrize to make it more generic, here is my proposal:

```C#
private static DictionaryVariable ensureVariableDictionaryEntryExists(CadDocument doc, string name, string value)
```

DomCR:
Merged! good job!

mme1950:
Thank you!

PR #805, opened 2025-09-23
Fix #

mme1950:
- Remove `Hatch.BoundaryPath.Ellipse.ICounterclockwise`
-  Read and use `CounterClockWise` instead

@DomCR This is a breaking change, please set the version info.


DomCR:
> This is a breaking change, please set the version info.

The version has already been updated to 1 in major, there is no release done yet, so I think we can include it as a minor.

Thanks for the heads up.

DomCR (approved):

PR #875, opened 2025-11-15
Issue 873 arcs and circles with radius 0 dxf

DomCR:
# Description

Fix for arcs and circles with radius 0 in dxf. The value will be assigned as Epsilon.


lawrence-laz:
Is there a planned date for the next nuget.org release?

PR #902, opened 2025-11-27
issue 901

DomCR:
# Description

Added the flag `IsUnloaded` for xref blocks.


DomCR:
Forgot to assign the flag in the template, apologies. Here is the PR with the fix:

- https://github.com/DomCR/ACadSharp/pull/906

PR #929, opened 2025-12-30
issue-928 TRACE entity for dxf

DomCR:
# Description

Implementation for TRACE entity for dxf.

# Notes for reviewer

Accidentally committed to master: https://github.com/DomCR/ACadSharp/commit/1ab772a3f683faf3becdc70bb26461c371e1a37e

TRACE entity has been deprecated, all the entities of this type will be written as SOLID instead which is the current equivalent with the same properties.


ilCosmico:
@DomCR I’m also interested in this fix. Do you have an ETA for when the updated NuGet package will be available?

In the meantime, Happy New Year!

PR #943, opened 2026-01-09
Fixes #942 - Added ToDegrees() to handle angular string representation

### LINKED ISSUE #942: Dimension style DecimalDegrees GetMeasurementText() not interpreting values correctly.
**Describe the bug**
I have a dwg with an angular dimension in degrees.
When using GetMeasurementText() I get the value in radians although the style has AngularUnitFormat.DecimalDegrees.

**To Reproduce**
Just load a *.dwg with an angular dimension.

**Expected behavior**
The value should be converted into degrees when the unit format is DecimalDegrees

**Screenshots**
**Additional context**
I'm not sure if the current UnitTest is actually correct:
```
new DimensionStyle {
				AngularUnit = AngularUnitFormat.DecimalDegrees
},
new DimensionAngular2Line{
				FirstPoint = XYZ.Zero,
				SecondPoint = new XYZ(10, 0, 0),
				AngleVertex = XYZ.Zero,
				DefinitionPoint = new XYZ(0,10, 0),
},
"1.57"
```
Shouldn't the resulting value be in degrees?
Taking a look at ezdxf for reference I understand this should be degrees:
https://ezdxf.readthedocs.io/en/stable/tutorials/angular_dimension.html

---

**Release needed**

- [ ] Yes, using the Nuget package.
- [x] No, using the github repo.

cpetry:
- corrected default AngularZeroHandling
- adjusted GetZeroHandlingFormat to actually use AngularDecimalPlaces
- added/adjusted UnitTests

# Description

As explained in #942 I have the impression that angular degrees are incorrectly converted to string.

# Tasks done in this PR
* [x] An evil bug have been defeated. :D

#942 

# Notes for reviewer
 - Not sure about the default values or the default handling of zeros

### DIFF
--- a/src/ACadSharp/Types/Units/UnitStyleFormat.cs
+++ b/src/ACadSharp/Types/Units/UnitStyleFormat.cs
@@ -392,8 +392,7 @@ namespace ACadSharp.Types.Units
 		/// Converts an angle value in degrees string representation.
 		/// </summary>
 		/// <param name="angle">Angle value in radians.</param>
-		/// <returns></returns>
-		/// <exception cref="ArgumentNullException"></exception>
+		/// <returns>A string that represents the value in degrees</returns>
 		public string ToDegrees(double angle)
 		{
 			double degrees = MathHelper.RadToDeg(angle);
@@ -408,8 +407,7 @@ namespace ACadSharp.Types.Units
 		/// Converts an angle value in degrees into its degrees, minutes and seconds string representation.
 		/// </summary>
 		/// <param name="angle">Angle value in radians.</param>
-		/// <returns></returns>
-		/// <exception cref="ArgumentNullException"></exception>
+		/// <returns>A string that represents the value in degrees, minutes and seconds</returns>
 		public string ToDegreesMinutesSeconds(double angle)
 		{
 			double degrees = MathHelper.RadToDeg(angle);


src/ACadSharp/Types/Units/UnitStyleFormat.cs
```diff
@@ -386,6 +388,22 @@ public string ToDecimal(double value, bool isAngular = false)
 			return value.ToString(this.GetZeroHandlingFormat(isAngular), numberFormat);
 		}
 
+		/// <summary>
+		/// Converts an angle value in degrees string representation.
+		/// </summary>
+		/// <param name="angle">Angle value in radians.</param>
+		/// <returns></returns>
+		/// <exception cref="ArgumentNullException"></exception>
```
DomCR: Where can the exception `ArgumentNullException` be thrown in the method?

I think I made the same mistake in `ToDegreesMinutesSeconds(double angle)` if there is no exception thrown, we can remove the documentation for this one in both cases.

DomCR:
The PR looks great! I left a couple of comments, once fixed, it can be merged.

Good job!

src/ACadSharp/Types/Units/UnitStyleFormat.cs
cpetry: You are totally right!
Copy&paste error reusing comment from ToDegreesMinutesSeconds.
Updated the PR

Issue #221, opened 2023-12-06
Aec architecture objects (wall, doors, windows)


cpetry:
**Is your feature request related to a problem? Please describe.**
When reading a DWG file, aec objects seem not to be included.

**Describe the solution you'd like**
I would like to have either AEC objects directly as entities or some way to access and read attributes (width, height, length, justification etc.).

**Additional context**
Aec objects are AutoCAD Architecture specific, I guess

DomCR:
I've been looking into this and there are a few problems that makes the implementation quite difficult. 
This type of objects are not compatible with dxf which means that I do not have a way to "see" them outside Autocad, also the there is no documentation about the different properties or variables that they have or how are they stored, not even in the OpenDesign document that has the DWG specification in it.

The objects are found in the reader and identified with a `DxfClass` but to be able to store the geometric data it will require to check the bytes manually which it will require a huge amount of manual work to decode byte by byte.

By now this feature will not be implemented, I'll leave the issue open just in case that I can find some documentation on it.

cpetry:
I would actually like to try reading it out byte by byte.
Can you please give me a headstart on how/where to begin? A branch with some basic setup?
I tried already but couldn't wrap my head around your architecture completely.

DomCR:
Here you go https://github.com/DomCR/ACadSharp/pull/227

I've created a simple setup so you can see how the `ObjectReader` works.

First the reader needs to identify the `ObjectType` that is reading, in this case is done by a `DxfClass` that contains the object name, in the example I added the AEC_WALL, after that, you have to finish the implementation of the method `readAecWall()` which needs to check manually the bytes in the stream. For that you can use the Explore method, which it will return a dictionary with all the possible values that you can read and then you can manually check them with the values that you have in Autocad.

The problem with all this is that you won't be able to tell if the reader has the whole information of the entity, some value are stored in flags or have conditions in which are only read if a property have a specific value, but you may be able to get the information that you want.

Let me know if I can help in anything.

Good luck!

cpetry:
Great!
I see how for I can get and if I need to add bytes later on.
Thanks for the headstart!

cpetry:
Oh God you are a lifesaver!
I tried at first and got stuck after a few bytes.
Good to know you had more luck

rurban:
@cpetry: Can you also please upload the DXF conversion from this AecObjects.dwg? Conversion without the AEC arx only converts to proxy entities, without any hint to the entity or object fields.

[...]

PR #969, opened 2026-02-06
Issue 221 aec architecture objects

cpetry:
# Description
Two years later I tried again.

Created basic structure.
Found handles to wallStyle and cleanupGroup.
BinRecord handle not found.
Tried analyzing BinRecord but no success so far. Unsure if this is related at all.

# Tasks done in this PR
* [x] Merge to current main
* [x] Somewhat worked on this

#221 
# Notes for reviewer
 - Nothing really working yet but easier setup to continue


src/ACadSharp/Entities/AecObjects/Wall.cs
```diff
@@ -0,0 +1,110 @@
...
+		/// Start position of the wall (insertion point).
+		/// </summary>
+		public XYZ StartPoint { get; set; } = XYZ.Zero;
+
+		/// <summary>
+		/// End position of the wall.
+		/// </summary>
+		public XYZ EndPoint { get; set; } = XYZ.Zero;
+		public XYZ Normal { get; internal set; }
+		public double Height { get; internal set; }
```
DomCR: Normal and Height should be able to be modified by the user, the internal can be removed.

We can leave it for the other properties like style by now and come back later.

src/ACadSharp/IO/DWG/DwgStreamReaders/DwgObjectReader.cs
```diff
@@ -1026,6 +1027,20 @@ private CadTemplate readUnlistedType(short classNumber)
 
 			switch (c.DxfName)
 			{
+				case DxfFileToken.EntityAecWall:
+					template = this.readAecWall();
+					break;
+				case "ACAECBINRECORD":
+				case "AcAecBinRecord":
+				case DxfFileToken.ObjectBinRecord:  // Add this variant
```
DomCR: `ACAECBINRECORD` and `AcAecBinRecord` seems to be the c++ naming for the class, are these cases real? I mean, in comse cases the `DxfName` becomes the c++ name?

DomCR:
General review notes:

- Before merging the PR, the AEC objects will need to be ignored by the writers, dwg and dxf.
- I've noticed that you  have stored the remaining bytes for the different objects, I don't think it is necessary, those bytes will be there without any use.

cpetry:
I tried analyzing these remaining bytes for their content. Currently it is not usable without having wall positions, width or any other meaningful property.

PR #1006, opened 2026-03-23
Hatch, apply transform fix.

Mikolaj505:
Applying transform to boundary entities as well as to edges, when applying transform to the hatch.

For several files I tried to open, entities where containing some of the boundary, and while hatches being part of the insert, they were missing applying transform when "unpacking" them, messing up the hatches badly.

This simple change fixes that missing behaviour.


DomCR:
resolved

Issue #727, opened 2025-08-06
Support for BOM


domeitzinger:
Thanks for this great library! 

The issue I am facing is that I have to read the BOM defined in a DWG file (as well as the balloon references for the parts). This would seem to require the implementation of the following types (from what I can tell by the logs): `ACMPARTLIST`, `ACMPARTREF`, `ACMBALLOON`

I would really appreciate it if this was possible.

DomCR:
Hi @domeitzinger,

Thanks for the support!

First, it can be done, but, the entities that you describe are not documented anywhere, nor dxf or dwg, which means that they have to be investigated manually, for dxf is easy, just an ascii file gives away a lot of information for dwg is a little bit more tricky.

Do you have a file that I can use where this objects are present?

domeitzinger:
Sure, I could send you a DWG (as well as a PDF export for reference) per mail. Should I use the mail address in your profile?

olix-cn:
> Yep

Hello, how is the progress of this feature?

DomCR:
Hi @olix-cn,

Unfortunately I haven't found a stable solution for this one.

@cpetry  added some new exploration methods on a PR, I'll try to use those methods to see if I can get some information.

Sorry for the inconvenient I'll keep you post it.

domeitzinger:
I tried to implement the parsing of those objects myself, but I only understand some chunks of the data. Also I'm not sure how reliable my findings are, given that there are many bytes whose meaning I could not figure out. 

Here are some of my findings so far:

### General
* The main entity seems to start with a repeating 2-byte value (e.g. `B4 04 B4 04`)
* The entities appear in DXF files and are represented by blocks with names of e.g. `*S120`, `*S121`, etc. and consist of basic entities
* The sub-entities seem to start with the byte pattern `0C 00 00 00 16 00 00 00 00 00 00 C0`. This pattern also sometimes appears additionally with `12`, `14` and `17` in place of `16`. This could be some kind of type indicator. 
* The sub-entities seem to end with the byte pattern `0C 00 00 00 16 00 00 00 FF FF FF FF`

## ACMPARTREF
The PartRef appears to only store the center position and radius of the marker. With this information, the entire entity can be drawn. This entity is in ModelSpace.

<img width="96" height="96" alt="PartRef" src="https://github.com/user-attachments/assets/53d66d2d-0ad8-4541-a127-caa9f5d2cd81" />

```csharp
private CadTemplate readPartRef()
{
[...]

PR #1019, opened 2026-04-03
Add mechanical entities and proxy graphics

### LINKED ISSUE #1122: Did you have any plan to implement full support for ProxyEntity?
Thanks for your project, and I want to know the plan of ProxyEntity? Or could I contribute my PR if possible?
I have read about ezdxf implementation: https://github.com/mozman/ezdxf/blob/master/src/ezdxf/proxygraphic.py and https://github.com/mozman/ezdxf/blob/master/src/ezdxf/entities/acad_proxy_entity.py, so I want to port them into this project.

domeitzinger:
# Description

Adds support for Mechanical specific entities `ACMPARTLIST`, `ACMPARTREF` and `ACMBALLOON` (tested with file version AC1032). 

Also implements reading of all known proxy entity graphics. Most of them were tested with multiple DWG files, but I haven't encountered some entity types (those have the `TODO: VALIDATE` comment in their parsing method). 

Additionally added the ability to read `PaddedUnicodeString` (PUS) and `PaddedString` (PS) values.

# Tasks done in this PR
* [X] Something awesome.

- #727 
- Based on #1016


domeitzinger:
I updated the `readLWPolyline()` method in `DwgObjectReader` to call a new internal `readLWPolyline` method which allows the ProxyLwPolyline to be read more easily (as it shares the same parsing logic as regular LwPolyline entities). I am not sure if this really is the best way to do this though.

DomCR:
> I updated the readLWPolyline() method in DwgObjectReader to call a new internal readLWPolyline method which allows the ProxyLwPolyline to be read more easily (as it shares the same parsing logic as regular LwPolyline entities). I am not sure if this really is the best way to do this though.

I don't like having duplicated code, but I would separate the 2 implementations just in case in the future they have to be changed or modified, if there are more cases like this I'll look into how to implement in a better way.

DomCR:
Let me know if you need help or you are finished with the implementation so I can take a final review and merge it.

Thanks a lot for this contribution!

domeitzinger:
> I don't like having duplicated code, but I would separate the 2 implementations just in case in the future they have to be changed or modified, if there are more cases like this I'll look into how to implement in a better way.

Makes sense, I reverted the code back to the original in the `DwgObjectReader.cs` and copied the code over to the ProxyLwPolyline parsing mehod. At least in the proxy entities I have not seen any other entities which would have the same code as regular entities.

> Let me know if you need help or you are finished with the implementation so I can take a final review and merge it.
> 
> Thanks a lot for this contribution!

I think most of the parsing should already work fine, but for a few objects I simply have no DWG files to test against or do not have enough test plans to better understand the values.

## Untested / Lack of plans
These are types I could not find in any DWG plans or had very few with non-conclusive values. The parsing is implemented according to the ODA documentation, but I could not validate if the data is correct. This would require more reference plans to test against.
* **SUBENT_FILLON (20)**
    > Supposed to be 0 or 1, but I have sometimes seen the value to be 2. This probably does not really matter as it is apparently supposed to be a bool.
* **PUSH_MODELXFORM2 (30)**
    > ODA docs also contains unknown values
* **RAY (13)**
* **MESH (8)**
* **PUSH_CLIP (27)**
* **SUB_ENTITY_MATERIAL (34)**
* **SUB_ENTITY_MAPPER (35)**
* **UNKNOWN37 (37)**
    > I do not know what this is supposed to be, there is no documentation and I did not see this entity in any plan
* ... more additional undocumented proxy entities

Therefore I think the implementations should be done so far and it would only require some additional testing to make sure the relatively untested entities are correct and do not cause crashes. I sadly do not have any DWG files for that though.

Also the BoundingBox calculation for the mechanical objects is currently only the position of the entity. This seems a bit inaccurate and should probably cover the BoundingBox of all its child proxy entities. I cannot really think of a good way to implement this though...

DomCR:
I'll merge this one into a transfer branch and run my battery of test.

Thanks!

DomCR (approved):

PR #1042, opened 2026-04-22
Insert empty pages to fill gaps

domeitzinger:
# Description

While looking into reading the Prototype_1b section of the file, I noticed the size of the section buffer was too small. There was a large gap of about 118784 bytes. Those bytes are all supposed to be 0 and the missing byte count would fit exactly into 4 decompressed pages of the section map. 

This implements the logic of the comment above the offset variable stating: `If this start offset is smaller than the sum of the decompressed size of all previous pages, then this page is to be preceded by zero pages until this condition is met`. I tested this change for the `readFileHeaderAC18` method and the section buffer size was correct with this change in place.

Looking at the `readFileHeaderAC21` code, a similar gap filling approach was already implemented, but commented out. I was unable to test this, but it seems to also be required (assuming the same logic applies here).

# Tasks done in this PR
* [ ] Something awesome.
* [X] An evil bug have been defeated.

### DIFF
--- a/src/ACadSharp/IO/DwgReader.cs
+++ b/src/ACadSharp/IO/DwgReader.cs
@@ -895,6 +895,7 @@ public class DwgReader : CadReaderBase<DwgReaderConfiguration>
 			descriptor.Name = decompressedStream.ReadString(64).Split('\0')[0];
 
 			//Following this, the following (local) section page map data will be present
+			ulong decompressSizeCounter = 0;
 			for (int j = 0; j < descriptor.PageCount; ++j)
 			{
 				DwgLocalSectionMap localmap = new DwgLocalSectionMap();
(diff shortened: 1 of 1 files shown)


DomCR:
Hi @domeitzinger,

The code that you commented out was meant to be deleted due it doesn't have any logic or practical use in the reading system, I think that adding those empty pages do not add any value to the reader.

I don't know if I'm missing something, did you find a practical use for this case that we need to add the empty pages?

domeitzinger:
It seems strange that a huge amount of bytes all being 0 is required, but here is an example of me trying to read the Prototype_1b section of some DWG file. Without the change this is the sections:
LocalSectionOffset
Between [1] and [2], 118784 bytes are missing. The file header of the Prototype_1b expects the file size to be 196352 bytes, but with the missing bytes, the actual stream size is only 77568 bytes. All offsets, sizes, etc. always refer to the full size (including the zeros) and this would cause the stream to read past the end (as you do not really know where the zeros have been skipped and the missing zeros also cause misalignment of the data). With the padding, the MemoryStream contains all bytes expected by the Section without missing data:
grafik

For some reason the Prototype_1b has 10987 entries of which only 9 have actual values and the rest are empty objects. Those empty objects are in the middle of the entries with actual data. When reading the section of entries, it contains a few entry definitions at first, then 10975 empty entries and then the remaining entries. Without adding the empty pages, the 10975 empty entries are cut off somewhere in the middle and this causes misalignment of the remaining data. So there would only be ~1088,3 entries instead of the total 10987.

So, without adding those empty pages, parsing the sections with large portions filled with zeros is really hard. I do not know why there even are e.g. 10975 empty entries within the Prototype_1b data section, but they seem to be actually valid objects.

DomCR:
Makes sense, good job!

DomCR (approved):

PR #1043, opened 2026-04-24
refactor: use System.Convert where applicable in group code value conversion

gherards99:
# Description

This PR normalizes value conversion by using `System.Convert` where applicable, while preserving the original switch order and formatting.
When calling `SetDimensionOverride()`, some override style values caused issues because explicit casts were not always valid.
For example, `ArrowHead.SeparateBlocks` raised an error because an `Int16` was expected, while the provided value was correctly a `bool`.
I tested these changes with my use cases, and everything worked as expected.

# Tasks done in this PR
* [ ] Something awesome.


DomCR (approved):
Good one!

PR #1053, opened 2026-04-30
Update GroupCodeValue.cs

### LINKED ISSUE #1051: Should DXF group codes 280–289 be Int16 instead of Byte?
Hi @DomCR,

While using the SetDimensionOverride() method, I encountered a problem when trying to set the DIMTZIN value of a dimension.

The issue is that, at a certain point, the group code for it is set as Byte and, because of that, after passing through the switch in ExtendedDataRecord, it throws a System.NotSupportedException.

Would it be wrong, in your opinion, to change Byte to Int16?
```
if (code >= 280 && code <= 289)
	return GroupCodeValueType.Byte;
```
After making that change, my use case works fine, but I am not sure whether it was originally defined that way intentionally.

Thanks.

gherards99:
# Description

Short description about this PR

Changed the GroupCodeValueType of codes 280 to 289 from Byte to Int16.

# Tasks done in this PR
* [ ] Something awesome.

https://github.com/DomCR/ACadSharp/issues/1051

### DIFF
--- a/src/ACadSharp/GroupCodeValue.cs
+++ b/src/ACadSharp/GroupCodeValue.cs
@@ -85,7 +85,7 @@ namespace ACadSharp
 				return GroupCodeValueType.Int16;
 
 			if (code >= 280 && code <= 289)
-				return GroupCodeValueType.Byte;
+				return GroupCodeValueType.Int16;
 
 			if (code >= 290 && code <= 299)
 				return GroupCodeValueType.Bool;


DomCR:
The pipeline is faling, we cannot change the value to Int16 due an incompatibility with the dwg format, is failing at `DwgObjectReader.readXRecord`.

The solution should be to fix the assign of DIMTZIN in some other way.

DomCR:
Fixed in:
- https://github.com/DomCR/ACadSharp/pull/1057

gherards99:
Hi @DomCR ,

thanks for the fix.

PR #1069, opened 2026-05-13
Update Hatch.BoundaryPath.Spline.cs

simonedd:
Fixed control points add

# Description

Fixed out of range exception (this.ControlPoints is initialized as empty list)


DomCR (approved):

DomCR:
Nice one!

PR #1073, opened 2026-05-15
Added missing MEASUREMENT system header variable

simonedd:
I'm not sure if it's mandatory on the DXF side, but I would add it anyway.


DomCR (approved):

DomCR:
Good one!

Issue #995, opened 2026-03-15
DxfTable in Dxf and Dwg writers


cxt240:
**Is your feature request related to a problem? Please describe.**
I do see that the dxf/dwg tables are being read by their readers. However, in the writers, they are marked as unsupported. I was wondering if it were possible to implement the writes.

https://github.com/DomCR/ACadSharp/blob/master/src/ACadSharp/IO/DWG/DwgStreamWriters/DwgObjectWriter.cs#L261

https://github.com/DomCR/ACadSharp/blob/master/src/ACadSharp/IO/DXF/DxfStreamWriter/DxfSectionWriterBase.Entities.cs#L138

**Describe the solution you'd like**
Implement writing for the table entities

PR #1075, opened 2026-05-15
Expose public TableEntity.CellStyleOverride accessor and table-level cell-style defaults

ilCosmico:
## Summary

Exposes `TableEntity.CellStyleOverride` (and `MergedCellRanges`) as public properties so external callers can populate table level overrides such as the flow direction. Also pre populates the default cell style instances with their canonical Id / HasData and adds a `DefaultTableCellStyle` referenced by the `TableStyle.TableCellStyle` initializer.

## Why

When writing a table to DWG R2010+, the per table flow direction (up/down) lives on `TableEntity.Content.CellStyleOverride` in the `FlowDirectionBottomToTop` bit of `PropertyOverrideFlags` and `TableCellStylePropertyFlags`. ODA reads it from there, not from the shared `TableStyle.FlowDirection`. Until now `Content` was `internal`, so external writers had no way to set that override and the only workaround was creating a separate `TableStyle` per direction, which then breaks the round trip through ODA.

This change is what lets a downstream DWG writer round trip the table flow direction correctly through ODA's reader.

## Changes

- `TableEntity.cs`
  - `public TableStyle.CellStyle CellStyleOverride { get; }` (delegates to `Content.CellStyleOverride`)
  - `public List<CellRange> MergedCellRanges { get; }`
  - Parameterless and `BlockRecord` based constructors
- `TableStyle.CellStyle.cs`
  - Default cell style factories now set the canonical `Id` (1 title, 2 header, 3 data) and `HasData = true`
  - New `DefaultTableCellStyle` (Id = 4, name 'Table')
- `TableStyle.cs`
  - `TableCellStyle` initializer uses `DefaultTableCellStyle`

3 files changed, +62 -1.

## Possibly related

This may or may not overlap with #995 (request to implement Dxf/Dwg writers for table entities). Linking it just in case it ends up being relevant.

## Test plan

- [x] Local build passes for all target frameworks
- [x] Round trip: write a `CadDocument` containing two tables (one with `FlowDirection.Down`, one with `FlowDirection.Up`) via an external writer that consumes `CellStyleOverride`, then read it back with ODA's reader. Table direction matches in both cases, which it didn't before this change.

### DIFF
--- a/src/ACadSharp/Entities/TableEntity.cs
+++ b/src/ACadSharp/Entities/TableEntity.cs
@@ -2,6 +2,7 @@
 using ACadSharp.Objects;
 using ACadSharp.Tables;
 using CSMath;
+using System;
 using System.Collections.Generic;
 
 namespace ACadSharp.Entities;
@@ -110,6 +111,25 @@ public partial class TableEntity : Insert
 	[DxfCodeValue(DxfReferenceType.Handle, 343)]
 	internal BlockRecord TableBlock { get { return this.Block; } }
 
+	/// <summary>
+	/// Initializes a new instance of the <see cref="TableEntity"/> class.
+	/// </summary>
+	public TableEntity() : base()
+	{
+	}
+
+	/// <summary>
+	/// Initializes a new instance of the <see cref="TableEntity"/> class
+	/// bound to the given anonymous <see cref="BlockRecord"/> that will host the table cache.
+	/// </summary>
+	/// <param name="block">Anonymous block record used as the table's owning block (DXF 343).</param>
+	/// <exception cref="ArgumentNullException"></exception>
+	public TableEntity(BlockRecord block) : base()
+	{
+		if (block is null) throw new ArgumentNullException(nameof(block));
+		this.Block = block;
+	}
+
 	/// <inheritdoc/>
 	public override CadObject Clone()
 	{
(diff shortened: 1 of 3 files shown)


src/ACadSharp/Entities/TableEntity.cs
```diff
@@ -110,6 +111,25 @@ public TableStyle Style
...
+	{
+	}
+
+	/// <summary>
+	/// Initializes a new instance of the <see cref="TableEntity"/> class
+	/// bound to the given anonymous <see cref="BlockRecord"/> that will host the table cache.
+	/// </summary>
+	/// <param name="block">Anonymous block record used as the table's owning block (DXF 343).</param>
+	/// <exception cref="ArgumentNullException"></exception>
+	public TableEntity(BlockRecord block) : base()
```
DomCR: The `BlockRecord` in the table is autogenerated like the dimensions, it should not be set by the user.

This constructor should be removed.

DomCR:
The PR looks really good! 

Please remove the constructor commented above and it's good to go.

ilCosmico:
Done, the `TableEntity(BlockRecord)` constructor has been removed (and the now unused `using System;` along with it). Force pushed to the same branch.

ilCosmico:
I had to revert the removal of the `TableEntity(BlockRecord)` constructor. Here's what I'm running into when I try to follow the "auto generated like the dimensions" path:

* I drop the constructor and create the table with `new TableEntity()`.
* `Insert.Block` has an `internal` setter, so from a downstream assembly I can't assign the anonymous `BlockRecord` to the table.
* With `Block == null`, `Insert.AssignDocument` early-returns (Insert.cs line 430), so the document never registers an anonymous `*T` block, `writeCommonInsertData` emits a null handle for DXF 343, and the resulting DWG has no rendered table cache. Re-reading the file then explodes the table back into `Line` and `MultilineText` entities scattered in model space instead of giving me back two `TableEntity`s.

I might be missing the autogeneration step. On the read side I can see `CadDimensionTemplate.Build` / `CadInsertTemplate.Build` setting `Block` after parsing, but on the write side I can't find a place where ACadSharp creates an anonymous block for a table whose `Block` is null. Could you point me at it, or alternatively would a public setter (or a `SetAnonymousBlock(BlockRecord)` method on `Insert` / `TableEntity`) be acceptable? Happy to update the PR with whichever you prefer.

In the meantime I've kept the constructor in the branch since that's the only public surface we currently have to bind a table to its block.

DomCR:
The block generation for the table is not yet implemented.

We can leave your current implementation with the constructor and I'll remove it after.

DomCR (approved):

PR #1077, opened 2026-05-18
Fix #1074: default VPort.AmbientColor to color index 250

### LINKED ISSUE #1074: DWG TrueView crashes on Realistic / Conceptual visual style with ACadSharp-written DWG (missing VisualStyle objects)
DWG files produced by `ACadSharp.IO.DwgWriter` crash DWG TrueView 2026. Two scenarios, same crash:

1. **Richer document** (e.g. a couple of `TableEntity` instances with their anonymous block populated, layers and viewport): the file opens fine in *2D Wireframe* mode. *Shaded* also renders. As soon as the user picks *Realistic* or *Conceptual* from the visual style menu, TrueView crashes.
2. **Minimal document** (just a single `Line` in model space): TrueView crashes on load, before any visual style change.

From `dwgviewr.err`:

```
FATAL ERROR:  Unhandled Access Violation Reading 0x0010 Exception at BF4A1EA3h
```

Base address changes each run, the `0x0010` offset doesn't. Looks like a null pointer dereference.

In `DwgObjectWriter.skipEntry()`, `VisualStyle`, `Material` and `AcdbPlaceHolder` are explicitly skipped and there are no `writeVisualStyle` / `writeMaterial` methods. A DWG written by Teigha/ODA from the same `CadDocument` model contains:

* 24 default `VisualStyle` entries under `ACAD_VISUALSTYLE`
* 3 `Material` entries (`Global`, `ByLayer`, `ByBlock`) under `ACAD_MATERIAL`
* 1 `AcdbPlaceHolder`

ACadSharp output has none. TrueView looks up the active visual style by name, doesn't find it, and dies.

## To Reproduce

```csharp
using ACadSharp;
using ACadSharp.Entities;
using ACadSharp.IO;
using CSMath;

var doc = new CadDocument(ACadVersion.AC1027);
doc.Entities.Add(new Line
{
    StartPoint = new XYZ(0, 0, 0),
    EndPoint   = new XYZ(100, 50, 0),
});

using var writer = new DwgWriter("minimal.dwg", doc);
writer.Write();
```

Open `minimal.dwg` in DWG TrueView 2026. Crashes on load, writes `dwgviewr.err` with the access violation above. This is scenario 2.

For scenario 1, take any larger document (a few `TableEntity` instances with a populated block is enough). The file opens in *2D Wireframe*. Pick *Realistic* or *Conceptual* from the visual style menu. TrueView crashes with the same access violation, which is when it tries to look up the style by name.

Same `CadDocument` model written through Teigha/ODA opens fine and switches visual styles fine.

### Diagnostic dump

Reading both DWGs back with `ACadSharp.IO.DwgReader.Read` and counting `NonGraphicalObject` types in the root dictionary tree:

| Object type      | ODA DWG | ACadSharp DWG |
|------------------|---------|---------------|
| AcdbPlaceHolder  | 1       | 0             |
| Material         | 3       | 0             |
| VisualStyle      | 24      | 0             |
| CadDictionary    | 15      | 15            |
| Scale            | 17      | 17            |
| TableStyle       | 1       | 1             |

## Expected behavior

DWG output should include at least the default `VisualStyle` (24) and `Material` (3) entries so AutoCAD and DWG TrueView can resolve handles when applying a visual style.

## Additional context

Reproduced with `ACadSharp 3.5.7` (NuGet), `ACadVersion.AC1027` (R2013).

The reader already has stubs for these objects. `DwgObjectReader.readVisualStyle` even carries a `TODO: Finish dwg implementation for VisualStyle` comment and only parses the header bits.

I tried the obvious shortcut: pull `Material` and `VisualStyle` off `skipEntry`, mirror the partial reader with a header-only writer. That corrupts the byte stream. TrueView reports 54 errors on recover and erases 29 objects. So a header-only emit isn't enough, the full per-object binary layout needs to be implemented.

## Sample repro project

[RepoVisualStyleCrash.zip](https://github.com/user-attachments/files/27805324/RepoVisualStyleCrash.zip)
Repro is also one paste of the snippet above.

**Release needed:** any.

* [ ] Yes, using the Nuget package.
* [x] No, using the github repo.

ilCosmico:
## Summary

Fixes #1074. Switching to Realistic or Conceptual on a DWG written by `DwgWriter` crashes DWG TrueView with `Unhandled Access Violation Reading 0x0010`.

## Root cause

`VPort.AmbientColor` is an auto-property with no initializer:

```csharp
public Color AmbientColor { get; set; }
```

`default(Color)` encodes as ByBlock. The `*Active` viewport has no block context, so the color cannot be resolved. When TrueView's Realistic or Conceptual render setup reads ambient color from the active VPORT it follows a null block pointer and aborts.

A reference AC1027 file produced by AutoCAD initializes `*Active.AmbientColor` to black. Reading that reference back through ACadSharp's own `DwgReader` and diffing the in-memory `VPort` against a fresh `new CadDocument(ACadVersion.AC1027)` pointed to this property as the only relevant difference.

## Fix

One line in `src/ACadSharp/Tables/VPort.cs`:

```csharp
public Color AmbientColor { get; set; } = Color.Black;
```

## Test plan

- [x] Built ACadSharp with the change.
- [x] Reproduced the issue from #1074 (single `Line`, AC1027) against the un-patched library; TrueView crashes on Realistic/Conceptual.
- [x] Same repro on the patched build; TrueView opens the file and switches visual style without crashing.
- [x] `DwgWriter` then `DwgReader` roundtrip of the patched output still succeeds.
- [x] Diff vs `master` is a single file, +1 / -1 lines.

### DIFF
--- a/src/ACadSharp/Tables/VPort.cs
+++ b/src/ACadSharp/Tables/VPort.cs
@@ -285,7 +285,7 @@ namespace ACadSharp.Tables
 		/// Ambient color(only output when non-black)
 		/// </summary>
 		[DxfCodeValue(63, 421, 431)]
-		public Color AmbientColor { get; set; }
+		public Color AmbientColor { get; set; } = Color.Black;
 
 		private XYZ _direction = XYZ.AxisZ;
 


src/ACadSharp/Tables/VPort.cs
```diff
@@ -285,7 +285,11 @@ public XYZ Direction
 		/// Ambient color(only output when non-black)
 		/// </summary>
 		[DxfCodeValue(63, 421, 431)]
-		public Color AmbientColor { get; set; }
+		// Default 250 matches AutoCAD/ODA. ByBlock (default(Color)) crashes
+		// DWG TrueView when applying Realistic/Conceptual visual styles because
+		// the viewport has no block context to resolve the ambient color from.
+		// See issue #1074.
+		public Color AmbientColor { get; set; } = new Color((short)250);
```
DomCR: Makes sense, please remove the comment and use `Color.Black` to assign the value.
Thanks

ilCosmico: Done, switched to Color.Black. Thanks.

DomCR (approved):

PR #1079, opened 2026-05-19
Fix DimensionDiameter point order in DWG reader/writer

simonedd:
Adjusted `10-pt` (DefinitionPoint) and `15-pt` (AngleVertex) order in DWG read/write as per ODA specification


DomCR (approved):

DomCR:
Good eye on this one, thanks!

Issue #996, opened 2026-03-16
Dimension Style Override Values in DxfWriter


gherards99:
Hi @DomCR,

I would like to ask if there is an easy way to set the various override style values of a dimension, such as Style.ArrowSize.

At the moment, the only working approach I have found is to create a new list of ExtendedDataRecord, add an ExtendedDataString called "DSTYLE" at the beginning, then add the ExtendedDataControlString entries to open and close {}, and finally place every override I need inside.

Is this currently the best way, or am I missing something?

Thanks.

PR #1081, opened 2026-05-19
Fix/dimension leader style override

### LINKED ISSUE #997: Set Leader Override
> Hi @DomCR,
> 
> At the moment, the SetDimensionOverride(DimensionStyle styleOverride) method is working great. I would like to ask what the correct approach would be for Leader entities instead, for example when setting Leader.Style.ArrowSize, because i can not see the same method there.
> 
> Thanks. 

 _Originally posted by @gherards99 in [#996](https://github.com/DomCR/ACadSharp/issues/996#issuecomment-4082761054)_

gherards99:
# Description

Three related fixes to make arrowhead style overrides work correctly in DXF output:

1. DimensionStyle.cs — SeparateArrowBlocks (code 173) defaulted to `true`, but the DXF standard default is 0 (false).

2. Dimension.cs — SetDimensionOverride skipped a property whenever *either* curr or over was null (`curr == null || over == null → continue`).  Handle-reference properties (ArrowBlock/DimArrow1/DimArrow2, codes 342-344) default to null, so assigning a BlockRecord in the override was silently ignored.

3. Leader.cs — Leader has a DimensionStyle Style property and uses the same ACAD/DSTYLE XDATA mechanism as Dimension, but had no SetDimensionOverride method.  The method (plus the supporting SetStyleOverrideMap) is added with identical diff-detection logic, using the same null-guard fix from point 2.  This allows callers to set e.g. overrideStyle.LeaderArrow = block and have DIMLDRBLK (code 341) written correctly.
This maybe could be refactored considering that is equal to the one for the Dimensions.

# Tasks done in this PR
* [ ] Something awesome.

# Notes for reviewer

### DIFF
--- a/src/ACadSharp/Entities/Dimension.cs
+++ b/src/ACadSharp/Entities/Dimension.cs
@@ -485,12 +485,16 @@ namespace ACadSharp.Entities
 				var curr = item.Value.GetRawValue(this.Style);
 				var over = item.Value.GetRawValue(styleOverride);
 
-				if (curr == null || over == null)
+				// Skip only when both values are null (no change to detect).
+				// When exactly one is null the property value has changed
+				// (e.g. a handle property going from null/default to a BlockRecord,
+				// or being explicitly cleared), so it must be included in the override.
+				if (curr == null && over == null)
 				{
 					continue;
 				}
 
-				if (!curr.Equals(over))
+				if (curr == null || over == null || !curr.Equals(over))
 				{
 					item.Value.StoredValue = over;
 					overrideMap.DxfProperties.Add(item.Key, item.Value);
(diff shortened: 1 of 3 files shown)


gherards99:
Related to issue #997 and issue #996 .

DomCR (approved):

DomCR:
Looks good! thanks for the contribution.

Sorry for the delayed review.

gherards99:
No worries, thank you as well.

gherards99:
Hi @DomCR,

sorry to bother you, but I was wondering if you have an estimate for when the next version of the NuGet packages will be released.

Thanks.

DomCR:
Hi @gherards99,

I'll make a release as soon as I can.

PR #1084, opened 2026-05-20
Add Material write support and Mesh texture coordinates round-trip

ilCosmico:
## Summary

Adds Material write support and Mesh texture coordinate round-trip in DWG.

- DwgWriter emits Material objects (writeMaterial body, dispatch, dropped from skipEntry).
- Entities persist their Material reference via the common entity data material flag bits, and CadEntityTemplate binds the handle on read.
- The override-color RGB packing in writeMaterial uses the 0xC2 tag in the high byte so AutoCAD reads it as a true-color override instead of falling back to ByLayer.
- Material render-mode hints are now strongly-typed enums: MaterialChannelFlags, MaterialIlluminationModel, MaterialMode replace the raw int fields.
- New Mesh.TextureCoordinates property persists per-vertex UVs through the ADSK_XREC_SUBDVERTEXTEXCOORDS XRecord that AcDbSubDMesh stores on the mesh extension dictionary. DwgWriter materializes the XRecord in prepareDocument before writeHeader so HandleSeed stays accurate.
- CadDictionary.Add(string key, NonGraphicalObject value) syncs value.Name with the supplied key when the value has no name yet. Without this, callers using the 2-arg overload lost the entry key on write because writers serialize item.Name as the dictionary entry name.
- Material.ChannelFlags default lowered from UseAll to UseDiffuse. UseAll caused DWG TrueView to drop texture rendering on basic materials because the renderer sampled channels (bump, reflection, refraction, normal) the material had no data for.

## Test plan

- [x] Diagnostic project writes 3 cubes with distinct Materials and reads them back: 3 materials, 3 meshes with material references, identical color/method/factor fields on read.
- [x] DWG TrueView 2025 opens a mesh exported via this stack in Realistic mode with the diffuse texture rendered correctly (textured house from an OBJ import).
- [x] Existing read path still parses files produced by the reference writer (verified by re-reading the same TrueView-friendly file through DwgReader after each change).

### DIFF
--- a/src/ACadSharp.Tests/Entities/MeshTests.cs
+++ b/src/ACadSharp.Tests/Entities/MeshTests.cs
@@ -1,0 +1,122 @@
+using ACadSharp.Entities;
+using ACadSharp.Extensions;
+using ACadSharp.IO;
+using ACadSharp.Tests.Common;
+using CSMath;
+using System.IO;
+using System.Linq;
+using Xunit;
+
+namespace ACadSharp.Tests.Entities
+{
+	public class MeshTests : CommonEntityTests<Mesh>
+	{
+		[Fact]
+		public override void GetBoundingBoxTest()
+		{
+			Mesh mesh = new Mesh();
+			mesh.Vertices.Add(new XYZ(-2, -3, -4));
+			mesh.Vertices.Add(new XYZ(5, 6, 7));
+			mesh.Vertices.Add(new XYZ(1, 0, 2));
+
+			BoundingBox boundingBox = mesh.GetBoundingBox();
+
+			Assert.Equal(new XYZ(-2, -3, -4), boundingBox.Min);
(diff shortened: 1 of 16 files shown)


ilCosmico:
Hi @DomCR, hope the table refactor is going well. Whenever you have a window, could you take a look at this one? It has been green since I opened it on May 20 and does not overlap the table work, so it should be a quick read. No rush, just flagging it so it does not get lost. Thanks as always for maintaining the project.

DomCR:
Hi @ilCosmico,

Sorry to keep this PR waiting but when a PR touches the writer I need to run manual test to ensure that everything is ok.

Yesterday I made a release so I'll transfer this to a branch and make a full review to merge it to master.

Thanks for the contribution!

DomCR (approved):

PR #1085, opened 2026-05-20
Add TableEntity DXF round-trip (cells, alignment, merges, flow direction)

### LINKED ISSUE #995: DxfTable in Dxf and Dwg writers

ilCosmico:
Until now the DXF writer only emitted the block reference for a TableEntity. This change writes the AcDbTable subclass too, with rows, columns, per-cell data, margins, flow direction, alignment (group 170 plus the matching cell flag at group 91), and merge ranges (per-cell via BorderWidth/BorderHeight and MergedValue).

On the reader side, per-cell alignment is now captured and the table-level MergedCellRanges is rebuilt by scanning the grid, so a DXF table comes back with the same merge layout as one read from DWG.

Known limitation: with no proxy graphic preview attached, AutoCAD and DWG TrueView draw the table twice (once from the anonymous block, once from the cell stream). Readers that consume the cell stream directly show a single, correct table.

Closes #995

### DIFF
--- a/src/ACadSharp/IO/DXF/DxfStreamReader/DxfSectionReaderBase.cs
+++ b/src/ACadSharp/IO/DXF/DxfStreamReader/DxfSectionReaderBase.cs
@@ -543,7 +543,18 @@ namespace ACadSharp.IO.DXF
 					tmp.CurrentCell.Rotation = this._reader.ValueAsDouble;
 					return true;
 				case 170:
-					//Has data flag
+					// Per-cell alignment (set when the cell flag at code 91 has the
+					// CellAlignment override bit). Capture it on the cell-level
+					// StyleOverride so callers reading the table back via DXF can
+					// resolve alignment exactly like they do through DWG.
+					if (tmp.CurrentCell != null)
+					{
+						tmp.CurrentCell.StyleOverride.HasData = true;
+						tmp.CurrentCell.StyleOverride.CellAlignment =
+							(ACadSharp.Objects.TableStyle.CellAlignmentType)this._reader.ValueAsShort;
+						tmp.CurrentCell.StyleOverride.PropertyOverrideFlags |=
+							ACadSharp.Objects.TableStyle.CellStylePropertyFlags.Alignment;
+					}
 					return true;
 				case 171:
 					tmp.CreateCell((TableEntity.CellType)this._reader.ValueAsInt);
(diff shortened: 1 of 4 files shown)


ilCosmico:
Hi @DomCR, I just saw your March comment on #995 about wanting to refactor CadTable before tackling the writers. This PR sits on top of the existing TableEntity API and does not touch CadTable internals, just the DXF reader and writer code paths, so it should not get in the way of that refactor. We had a downstream need for the round trip sooner, which is why I went ahead. If your refactor lands first and the shape changes, I can rebase, or close this PR if you would rather start fresh. Whatever works best for you.

ilCosmico:
Hi @DomCR, coming back to this now that the CadTable refactor has landed. The PR still merges cleanly on top of the new API, and it covers the TableEntity DXF write path that is currently a not-implemented notify on master.

I saw you have already started on the writers with the TableStyle DXF support, so I did not want to just push ahead and step on that. If it helps, I am glad to lend a hand finishing the table writers starting from this PR, rebasing or reshaping it to match your refactor, and taking the DWG side too if you want it. Totally your call on scope and how to split things. And if you would rather keep this part yourself, no problem at all, just let me know and I will step back.

Either way, thanks for the refactor, it made this much cleaner to build on.

DomCR:
Hi @ilCosmico,

Before implementing the support for the tables in dxf I want to solve the problem with the `TableContent` class, which is obsolete for the newer versions of dwg but is used in older dwg and dxf. 

I still need to refactor the dxf reader to fix the mess that the `TableContent` created and see how can it be created if the user needs to write it.

Having said that, if this implementation works with existing files and it rewrites a dxf file with a table, I can create a transfer branch and check it once I refactor the `TableContent` part.

Let me know what you prefer to do.

ilCosmico:
Hi @DomCR, that plan works for me. The transfer branch is fine, and I will rebase or adapt the PR after the TableContent refactor lands, whatever shape it takes.

On your question: yes, the implementation round-trips existing files. A dxf with a table reads and rewrites with cells, per-cell alignment, merge ranges and flow direction preserved, and we validated it downstream with our product tests against files from real drawings. One known limitation, noted here for transparency: the writer does not emit the proxy graphic preview blob, so viewers regenerate the table from the data instead of the cached preview. Content and layout are correct, only the hover preview behaves differently, as in the screenshot below (the rotated overlay is the regenerated preview during hover).

A_TrueView

Take your time with the refactor, and thanks for keeping this on the table 😅

ilCosmico:
Rebased on current master, plus two commits. Cell values are now written for every value type found in the DXF that AutoCAD saved from `sample_AC1032.dwg` (long, double, string, date and point), with their flags, units and format taken from the model. Before, only text values were written, so numbers and dates were lost. FIELD values go through the same code and now read back as well. I also reworded a few comments.

On `sample_AC1032.dwg` all 19 cells come back with their type and value, and DWG TrueView shows them with their formats. To open that file in TrueView I had to work around the hatch issue reported in #1253, which is unrelated to tables.

PR #1086, opened 2026-05-21
Updated check on default layouts removing

simonedd:
# Description

`LayoutCollection.Remove` blocked "Layout1" by name, but the name has no structural meaning in DWG/DXF. Only the count of paper layouts matters. Replaced the name check with a count check.

An alternative would be to remove the layout and recreate a default 'Layout1' if it is the last one. Let me know If you prefer this solution.

### DIFF
--- a/src/ACadSharp/Objects/Collections/LayoutCollection.cs
+++ b/src/ACadSharp/Objects/Collections/LayoutCollection.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Linq;
 
 namespace ACadSharp.Objects.Collections;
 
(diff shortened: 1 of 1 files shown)


DomCR:
The solution makes sense, and fixes the issue.

Have you tested this scenario:
- Create a CadDocument with Model, Layout1 (default, with the block record assigned to *Paper_Space) and Layout2
- Remove Layout1 keeping Layout2
- Save the document
- Can it be opened?

If the scenario works fine this one can be merge.

Thanks for the contribution!

simonedd:
You're right, this scenario doesn't work. I investigate a bit deeper and is more complex than expected. 
I think the Layouts management should follow the following rules:

**Naming convention**

- The active paper layout's BlockRecord is named *Paper_Space (no suffix).
- Inactive paper layouts use *Paper_Space0, *Paper_Space1, … (smallest free index, non-consecutive tolerated).
- The Layout's own Name ("Model", "Layout1", "Sheet1", …) is independent from the BlockRecord name.
- The Model layout is always tied to the *Model_Space BlockRecord.

**The collection must enforce**

- Exactly one *Model_Space BlockRecord ↔ one Model Layout. Neither can be removed.
- At least one paper layout must exist at any time (so exactly one *Paper_Space BlockRecord).
- No orphan paper-space BlockRecords (every *Paper_Space* BR has a Layout pointing to it, and vice versa).
- Add(Layout) — when a paper layout is added to a document, its AssociatedBlock is automatically Added/Named
- Remove(layoutName) — if removing the active paper layout, promote another remaining paper layout to active. Always remove the layout's AssociatedBlock from the document so no orphan paper-space BR survives.

**Difference with current behaviour**
- Layout.Name indipendent from AssociatedBlock name and the Layout.AssociatedBlock must be created only when the Layout is added to a Document
- The BlockRecord must be removed from the BlockRecordsTable if the associated Layout is removed

What do you think?

DomCR:
The method of "promotion" is quite complicated to implement.

I think that we can go for a more simple approach, which is to check the block that is attached to the layout, if the block is the model or the main paper space, it should not be allowed to be removed.

It would not allow the user to remove the Layout linked to the *Paper_Space block but it would simplify the operation and ensure integrity.

Another way to do this operation is not allow the user to delete any layout but the model one, and before the writing operation add the default paper layout (if not present) and rename the layout blocks. This approach is quite more invasive and changes variables (block names) that the user may be using somewhere else so I'm not quite convinced with this one.

simonedd:
Ok. I'll try to go for the simple approach.
I'm closing this PR in favor of a proper branch-based PR.

PR #1096, opened 2026-05-26
Improved Paper Space Layout management

simonedd:
# Description

Follow-up of PR #1086

- When removing a layout, the check on default layout is based on the BlockRecord name and the AssociatedBlock is removed from the Document.BlockRecords collection.
- When adding a layout, the default _*Paper_SpaceN_ block name is assigned to the AssociatedBlock.

### DIFF
--- a/src/ACadSharp/Objects/Collections/LayoutCollection.cs
+++ b/src/ACadSharp/Objects/Collections/LayoutCollection.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Linq;
+using ACadSharp.Tables;
 
 namespace ACadSharp.Objects.Collections;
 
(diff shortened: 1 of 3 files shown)


src/ACadSharp/Objects/Layout.cs
```diff
@@ -311,9 +311,8 @@ internal override void UnassignDocument()
 
 		if (this.AssociatedBlock != null)
 		{
+			this.Document.BlockRecords.Remove(this.AssociatedBlock.Name);
 			this.AssociatedBlock.Layout = null;
-			this.Document.BlockRecords.OnRemove -= this.onRemoveBlockRecord;
-			this._blockRecord = (BlockRecord)this._blockRecord?.Clone();
```
DomCR, line 316: We cannot remove the line:
```C#
this.Document.BlockRecords.OnRemove -= this.onRemoveBlockRecord;
```
If this is removed the layout will act when the document removes a block, which may cause a malfunction.

src/ACadSharp/Objects/Collections/LayoutCollection.cs
```diff
@@ -18,15 +20,44 @@ public LayoutCollection(CadDictionary dictionary) : base(dictionary)
 		this._dictionary = dictionary;
 	}
 
-	/// <inheritdoc/>
+	public override void Add(Layout entry)
+	{
+		if (!isDefaultLayout(entry))
+			entry.AssociatedBlock.Name = findSmallestFreePaperSpaceName(this._dictionary.Document);
```
DomCR, line 26: I don't think is necessary to change the name of the block, I know that the standard has to be named PaperSpace(number) but after the first one, the others don't seem to effect the output.

src/ACadSharp/Objects/Collections/LayoutCollection.cs
```diff
@@ -18,15 +20,44 @@ public LayoutCollection(CadDictionary dictionary) : base(dictionary)
...
 		return base.Remove(name, out entry);
 	}
+
+	private static bool isDefaultLayout(Layout layout)
+	{
+		return layout.AssociatedBlock.Name.Equals(BlockRecord.PaperSpaceName, StringComparison.InvariantCultureIgnoreCase) ||
+		       layout.AssociatedBlock.Name.Equals(BlockRecord.ModelSpaceName, StringComparison.InvariantCultureIgnoreCase);
+	}
+
+	private static string findSmallestFreePaperSpaceName(CadDocument doc)
```
DomCR, line 50: This method could be a public method for the user to assign the next valid name to the layout block.

Another option could be to implement the method `CreateLayout` in the collection which it will add a layout with the valid name and return the result.

DomCR:
Apologize for the delayed review.

src/ACadSharp/Objects/Layout.cs
simonedd: It was a duplicated line. The unsubscribe from the event is the first line of the method.

src/ACadSharp/Objects/Collections/LayoutCollection.cs
simonedd: Yes, it's not required. DWG files are still valid, though.

simonedd:
While working on this I realized that proper paper-space layout management is a larger, interconnected task than what this PR covers. Should probably be done as one coherent effort. 
I've also solved my underlying issue without needing changes to ACadSharp, so there's no urgency on my side.

Is there anything here you'd want to keep? Otherwise I'm happy to close it. Let me know what you prefer.

DomCR:
I agree that the Layouts have a tight coupling with the Associated block which causes a rise of complexity we can close this one and I'll take a look on that relationship between blocks and layouts.

Thanks for the help!

PR #1097, opened 2026-05-26
Add DWG reading for BLOCKLINEARPARAMETER

mdelanno:
## Summary

Adds DWG read support for the `BLOCKLINEARPARAMETER` object. The data-model class `BlockLinearParameter` already existed, but it was never read by `DwgObjectReader` (no `case` and no reader method), so its values were lost
when loading a drawing.

## Changes

- `BlockLinearParameter`: adds `BaseLocation`, `Minimum`, `Maximum`,  `Increment` and the `Values` list (the discrete allowed distances).
- New `LinearParameterBaseLocation` enum (start point / middle point).
- New `CadBlockLinearParameterTemplate` (mirrors `CadBlockRotationParameterTemplate`).
- `readBlockLinearParameter` in `DwgObjectReader.Objects.cs` + the matching  `case DxfFileToken.ObjectBlockLinearParameter` in the reader switch.

## Implementation note

`readBlockLinearParameter` reads the connection table with a fixed sequence instead of reusing `readBlock2PtParameter`, because the generic connection-list loop in `readBlock2PtParameter` was not byte-compatible with the blocks I tested. I kept the two readers separate to avoid changing the existing rotation path, but I'm happy to align `readBlockLinearParameter` with `readBlockRotationParameter` if you prefer the shared helper.

## Testing

Validated against a real dynamic-block library: the distance min/max, base location and discrete value list extracted from each linear parameter match a known-good reference exactly. No sample DWG is added here because the blocks are proprietary, but I can provide a minimal repro if that would help.

### DIFF
--- a/src/ACadSharp/IO/DWG/DwgStreamReaders/DwgObjectReader.Objects.cs
+++ b/src/ACadSharp/IO/DWG/DwgStreamReaders/DwgObjectReader.Objects.cs
@@ -215,6 +215,65 @@ internal partial class DwgObjectReader : DwgSectionIO
 		return template;
 	}
 
+	private CadTemplate readBlockLinearParameter()
+	{
+		BlockLinearParameter blockLinearParameter = new();
+		CadBlockLinearParameterTemplate template = new CadBlockLinearParameterTemplate(blockLinearParameter);
+
+		// The connection table is read inline with a fixed sequence instead of calling
+		// readBlock2PtParameter, whose generic connection-list loop is not byte-compatible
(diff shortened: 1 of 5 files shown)


DomCR:
Hi @mdelanno,

The tests are not passing, this needs another iteration.

DomCR (approved):

DomCR:
The implementation looks good, I'll merge it after a manual testing on my end and a file specific test.

Thanks for the contribution!

PR #1106, opened 2026-06-08
Fix GetActiveColor() to resolve nested ByLayer when Color IsByBlock in Entity.cs

TVIX:
- I believe this GetActiveColor() function is intended to return the actual color used to render an entity, not the ByLayer or ByBlock indicator.
- While using the function I received a ByLayer value back, which is not the expected actual color. During debugging I found that the IsByBlock branch returned a color that was ByLayer instead of a resolved actual color. To address that, I added the nested check so that when the block’s color resolves to ByLayer, we instead return the entity’s layer color.
- I am not sure this handles all nested or edge cases. There may be other combinations of nested ByBlock/ByLayer or other owner types that still return an indicator instead of an actual color.
- This tentative fix is based on my debugging and current understanding. It may need a more comprehensive resolution strategy (for example, a generic color resolution helper that traverses layers/blocks until an actual color is found).
- Please review.

### DIFF
--- a/src/ACadSharp/Entities/Entity.cs
+++ b/src/ACadSharp/Entities/Entity.cs
@@ -169,6 +169,10 @@ public abstract class Entity : CadObject, IEntity
 		else if (this.Color.IsByBlock && this.Owner is BlockRecord record)
 		{
 			color = record.BlockEntity.Color;
+			if (color.IsByLayer)
+			{
+				color = this.Layer.Color;
+			}
 		}
 		else
 		{


DomCR:
Hi @TVIX,

Your implementation seems correct, I'm not sure how it has to be resolved at the end but your current fix gives a more coherent resolution for the problem.

I can merge this one to have this specific edge case fix, I'll investigate how the resolution of the color has to work if is assigned to `IsByBlock` flag is on. The fix should take in account that the `BlockEntity` implements the `GetActiveColor` but it also overrides the `Color` property from the `Entity` to avoid the assignation of `ByBlock`.

Thanks for the contribution!

DomCR (approved):

TVIX:
I'm glad this turned out to be of some use for the project.

I’d like to add one more note that might help. I initially tried implementing a recursive call to `GetActiveColor()`, but that turned out to be incorrect. The issue is that a `BlockEntity` is typically placed on the default layer, not on the layer of the current `Entity (this)`. As a result, the recursion eventually resolves to a actual color, but it is the wrong one.

```
else if (this.Color.IsByBlock && this.Owner is BlockRecord record)
{
	color = record.BlockEntity.GetActiveColor();
}
```

> Hi @TVIX,
> 
> Your implementation seems correct, I'm not sure how it has to be resolved at the end but your current fix gives a more coherent resolution for the problem.
> 
> I can merge this one to have this specific edge case fix, I'll investigate how the resolution of the color has to work if is assigned to `IsByBlock` flag is on. The fix should take in account that the `BlockEntity` implements the `GetActiveColor` but it also overrides the `Color` property from the `Entity` to avoid the assignation of `ByBlock`.
> 
> Thanks for the contribution!

PR #1110, opened 2026-06-09
Fix insert transforms for block base points

chui-xelleron:
Insert.GetTransform() now subtracts the block base point before applying the insert transform, which fixes composite block placement when the block definition is offset from (0,0).

This PR includes:
- a focused regression test covering a non-zero block base point
- a small example showing the same case in the examples project

Verified with:
- dotnet test src/ACadSharp.Tests/ACadSharp.Tests.csproj --no-build -f net9.0 --filter FullyQualifiedName~ACadSharp.Tests.Entities.InsertTest
- dotnet build src/ACadSharp.Examples/ACadSharp.Examples.csproj


DomCR (approved):

DomCR:
Looks good! Thanks for the contribution!

PR #1121, opened 2026-06-22
Add AcDbBlockLookupParameter reader + expose 2pt property labels

ElieBouillard:
# Description

`AcDbBlockLookupParameter` objects were silently dropped during DWG parsing because no handler was registered for them in the object reader dispatch switch. This made it impossible to read the display name of Lookup-type parameters from dynamic blocks.

This PR adds minimal support for `BlockLookupParameter` and exposes `PropertyLabels` on `Block2PtParameter`, which together allow consumers to enumerate all parameters visible in AutoCAD's "Customize" tab for a dynamic block insert.

# Tasks done in this PR
* [x] Add `DxfFileToken.ObjectBlockLookupParameter` constant.
* [x] Add `BlockLookupParameter` model class (`Block1PtParameter` subclass) with a `Name` property (DXF code 301) holding the parameter display name.
* [x] Register `BLOCKLOOKUPPARAMETER` in the DWG object reader dispatch switch.
* [x] Add DXF reader support via `readBlockLookupParameter` in `DxfObjectsSectionReader`.
* [x] Expose `PropertyLabels` on `Block2PtParameter` (populated from DWG codes 303/304), allowing callers to identify parameters that publish named properties to the Properties palette.
* [x] Add `assertLookupParameter` test scaffolding and the corresponding `IsolatedTest` case in `DynamicBlockTests`.

- None.

# Notes for reviewer
- The lookup table body is intentionally not decoded — only the common block-element prefix and the display name are read. The object is deserialized from its own bounded stream so leaving the tail unread is safe.

### DIFF
--- a/src/ACadSharp.Tests/IO/DynamicBlockTests.cs
+++ b/src/ACadSharp.Tests/IO/DynamicBlockTests.cs
@@ -100,6 +100,9 @@ namespace ACadSharp.Tests.IO
 				case DxfFileToken.ObjectBlockLinearParameter:
 					this.assertLinearParameter(doc);
 					break;
+				case DxfFileToken.ObjectBlockLookupParameter:
+					this.assertLookupParameter(doc);
+					break;
 				default:
 					throw new System.NotImplementedException();
 			}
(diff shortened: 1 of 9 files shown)


DomCR (approved):

DomCR:
Hi @ElieBouillard,

Nice one! I'll create a transfer branch to test it on my end and merge it to master.

Thanks for the contribution!

Issue #1133, opened 2026-07-06
DWG/DXF writers: serialize EvaluationGraph + BlockRepresentationData (dynamic-block round-trip)


Sipowiz:
First: thank you for ACadSharp — the dynamic-block *reading* support (visibility states, linear/flip/rotation parameters, #161 / #948 / #962 / #1097) has been excellent for us.

We build document-management software for construction/renovation engineering (title-block and page-setup synchronization across large DWG sets). We would like to use ACadSharp for writing, but currently cannot for one reason: `DwgObjectWriter` (and the DXF object writer) skip `EvaluationGraph`, `BlockRepresentationData` and the related context data with `[NotImplemented]` notifications. The consequence is that any read-modify-write of a drawing that contains dynamic blocks (in practice: every title block and drawing frame in our customers' files) silently converts them to static anonymous blocks when reopened in AutoCAD. Attribute values, geometry and the `AcDbDynamicBlockTrueName` XData survive — but the parameters, grips and visibility states are gone, so we currently treat writing user files as off-limits.

We verified on 3.6.35 (and current master) with real production templates:

- Reading + mutating attributes / layouts / plot settings round-trips losslessly.
- The instance state in `AcDbBlockRepresentation` → `AppDataCache` → `ACAD_ENHANCEDBLOCKDATA` is readable (and even editable in memory), but is dropped by the writer along with the definition-side evaluation graph.
- Grip/action objects (`BLOCKLINEARGRIP`, `BLOCKSTRETCHACTION`, `BLOCKARRAYACTION`, `ACDB_DYNAMICBLOCKPURGEPREVENTER_VERSION`) are read as `UnknownNonGraphicalObject`, so they are lost as well.

Related: #893, #408, PR #993.

We have forked the repo and started work on this ourselves, targeting the scoped goal of **round-tripping what the reader already parses** (re-emit the evaluation graph / representation data read from the source file — not authoring new dynamic blocks from scratch). Questions before we get too deep:

1. Is there a design direction you'd want followed — e.g. mirroring the `DwgObjectReader` templates the way the Material writer was added in #1084?
2. Would you accept a PR scoped as above, possibly landed in stages (typed objects first, grip/action types after)?
3. Any known gaps in the read model we should be aware of before attempting this? (In #948 you mentioned the ENHANCEDBLOCK structures were mapped by exploration.)

We're prepared to invest engineering time in this — for us it is the difference between ACadSharp being a read-only library and a full solution.

[...]

PR #1136, opened 2026-07-06
Fix DXF objects-section writer: dangling handles, duplicated objects, null field text

Sipowiz:
Three independent robustness fixes for the DXF objects-section writer, found while round-tripping real production dynamic-block drawings (context: #1133). Each is a separate commit.

### 1. Dangling dictionary handles (`Fix dangling dictionary handles in the dxf writer`)
`DxfObjectsSectionWriter.writeDictionary` emitted the `3`/`350` entry-name/handle pair (and enqueued the entry) for **every** dictionary entry, even entries whose object type the writer later drops via `isObjectSupported` (`ACAD_ENHANCEDBLOCK`, `AcDbRepData`, visual styles, purge preventers…). The result is a DXF file with dictionary entries pointing at handles that were never written — re-reading it raises `Entry not found …` notifications. Fixed by filtering entries through `isObjectSupported` before emitting the pair, mirroring the DWG writer's `skipEntry` double-filter. Covered by two new tests (`DwgWriterNoDanglingDictionaryEntriesTest` / `DxfWriterNoDanglingDictionaryEntriesTest`) that rewrite the dynamic-block fixtures in memory, re-read, and assert no `Entry not found` notification is raised.

### 2. Duplicated objects in the objects section (`Avoid writing duplicated objects in the dxf objects section`)
An object reachable from two owners (e.g. a `Field` enqueued by both its owner dictionary and the `FieldList` that references it) was written twice into the OBJECTS section; re-reading the output then failed with `Item already has an owner`. Fixed by tracking written handles in a `HashSet<ulong>` and skipping an object already emitted. Reproducible with any FIELD-bearing drawing.

### 3. Null text in `writeLongTextValue` (`Handle null text values in writeLongTextValue`)
`DxfSectionWriterBase.writeLongTextValue` threw a `NullReferenceException` on `FIELD` objects whose cached value string is null. Fixed by treating null as empty, matching the DWG writer's behaviour.

Fixes 2 and 3 don't ship a dedicated fixture here because they were surfaced by proprietary drawings; both reproduce with any FIELD-bearing DXF and I'm happy to add a public fixture if you'd like one. All existing `DynamicBlockTests` pass (45/45) on top of current master (v3.6.37).

### DIFF
--- a/src/ACadSharp.Tests/IO/DynamicBlockTests.cs
+++ b/src/ACadSharp.Tests/IO/DynamicBlockTests.cs
@@ -5,8 +5,6 @@ using ACadSharp.Objects.Evaluations;
 using ACadSharp.Tables;
 using ACadSharp.Tests.TestModels;
 using System;
-using System.Collections.Generic;
-using System.IO;
 using System.Linq;
 using Xunit;
 using Xunit.Abstractions;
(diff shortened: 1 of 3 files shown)


src/ACadSharp.Tests/IO/DynamicBlockTests.cs
DomCR, line 1: In this PR there is no advance on writing the dynamic parameters, the changes in these tests are not needed by now.

DomCR (approved):

DomCR:
To keep the task on scope, please revert the changes made in src/ACadSharp.Tests/IO/DynamicBlockTests.cs and is good to go.

Sipowiz:
Done — reverted the `DynamicBlockTests.cs` changes, so this PR is now just the three DXF writer fixes. Thanks for the review! Happy to contribute the dangling-handle / dup-object regression tests separately in whatever form fits your test layout.

PR #1137, opened 2026-07-06
Write BlockRepresentationData (AcDbRepData) in DWG and DXF

Sipowiz:
First step toward the dynamic-block write support discussed in #1133: writing `BlockRepresentationData` (`AcDbRepData`) in both the DWG and DXF writers.

### What this does
`BlockRepresentationData` is the per-insert object (under `Insert.XDictionary → AcDbBlockRepresentation → AcDbRepData`) that links an anonymous `*U…` block instance back to its source dynamic block. Its sibling `AppDataCache → ACAD_ENHANCEDBLOCKDATA → XRecord` chain already round-trips through the existing dictionary/XRecord writers — only `AcDbRepData` itself was dropped by `skipEntry` / `isObjectSupported`.

- Removes `BlockRepresentationData` from the DWG `skipEntry` and DXF `isObjectSupported` drop-lists.
- `writeBlockRepresentationData` in both writers, mirroring the readers: `BS 70` (`Value70`) + a hard pointer (`340`) to the source `BlockRecord`.
- Adds the `AcDbBlockRepresentationData` `DxfClass` entry (values taken from AutoCAD-authored files: `ProxyFlags=1153`, `ItemClassId=499`, `DwgVersion=AC1018`, `MaintenanceVersion=58`; identical across the AC1027 and AC1032 files I checked).
- Refactors `DynamicBlockTests` asserts into a reusable `assertRepresentationChain` layer and adds unconditional DWG + DXF round-trip theories (they are **not** gated on `SELF_CHECK_OUTPUT`, so they run in CI).

### Scope / what this deliberately does NOT do
This does not write the `EvaluationGraph` (`ACAD_ENHANCEDBLOCK`) — you mentioned in #1133 you're exploring that yourself, so I've kept this PR to the representation-data half and left the graph alone.

### Validation against real AutoCAD
I ran the rewritten files through AutoCAD 2022 (via COM `AUDIT` + block-reference enumeration) on a corpus of real production dynamic-block templates: all open **without a recover prompt** and audit-clean apart from a finding that is already present in the untouched source file (an unrelated `ACAD_LAYERFILTERS` XRecord length). One honest caveat: restoring `AcDbRepData` alone does **not** make `IsDynamicBlock` return true — that needs the graph — but it adds no audit errors and is a prerequisite for it. 61/61 `DynamicBlockTests` pass.

### Note
Stacked on #1136 (the DXF objects-section writer fixes) — the round-trip tests here rely on those. I'll rebase once #1136 merges; until then this branch contains those three commits too.

### DIFF
--- a/src/ACadSharp.Tests/IO/DynamicBlockTests.cs
+++ b/src/ACadSharp.Tests/IO/DynamicBlockTests.cs
@@ -88,23 +88,7 @@ namespace ACadSharp.Tests.IO
 			var config = getConfiguration(test);
 			var doc = this.readDocument(test, config);
 
-			switch (test.NoExtensionName)
-			{
-				case DxfFileToken.ObjectBlockVisibilityParameter:
-					this.assertVisibilityParameter(doc);
-					break;
-				case DxfFileToken.ObjectBlockRotationParameter:
-					this.assertRotationParameter(doc);
-					break;
-				case DxfFileToken.ObjectBlockPointParameter:
-					this.assertPointParameter(doc);
-					break;
-				case DxfFileToken.ObjectBlockLinearParameter:
-					this.assertLinearParameter(doc);
-					break;
-				default:
-					throw new System.NotImplementedException();
-			}
+			this.assertIsolatedDocument(test, doc, assertEvaluationGraph: true);
 		}
 
 		[Theory]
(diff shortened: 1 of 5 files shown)


DomCR:
It seems that there is a duplication of the task done in:
- https://github.com/DomCR/ACadSharp/pull/1136

Sipowiz:
Sorry for the confusion — #1137 is **stacked on top of #1136**, so its branch currently contains #1136's three commits as well; that's the overlap you're seeing, not a re-do. Its own change is a single commit: the `BlockRepresentationData` (`AcDbRepData`) writer for DWG + DXF (`writeBlockRepresentationData` = BS 70 + hard pointer 340 to the source `BlockRecord`) plus the `AcDbBlockRepresentationData` `DxfClass` entry.

Once #1136 merges I'll rebase this onto master so the diff shows only the RepData writer. Want me to rebase now onto #1136's branch instead, or hold until #1136 lands? (I'll also drop the `DynamicBlockTests.cs` additions here to match the scoping you asked for on #1136.)

DomCR:
Hi @Sipowiz,

I've been working on the dynamic parameters and I managed to implement the first successful result for the DxfWriter:
- https://github.com/DomCR/ACadSharp/pull/1151

I think this one can be closed.

Issue #822, opened 2025-10-12
Cannot retrieve data from the region


zcfire2017:
I have a triangle composed of multiple line segments, and I set the triangle as a region. When I read it, I can get the data of the region, but I cannot get the data of multiple line segments under the region

<img width="803" height="437" alt="Image" src="https://github.com/user-attachments/assets/9af32393-454f-443a-a682-52339c889993" />

<img width="1759" height="529" alt="Image" src="https://github.com/user-attachments/assets/32b5f3ff-2f33-4ede-aa6e-a017ecb463ea" />

AKC-DuyLoc:
I am also having the same problem about region entities. Can you tell me where to get region information. Thank you!

AKC-DuyLoc:
@DomCR Thank you for this useful information

AKC-DuyLoc:
@DomCR In DwgReader, I see ACIS entities haven't handle in here. I think that is reason. I am applying to handle this issue. 

<img width="1072" height="519" alt="Image" src="https://github.com/user-attachments/assets/1f4ac462-1c07-433d-b061-0d5a682d2106" />

PR #1139, opened 2026-07-07
Read the ACIS payload of 3DSOLID, REGION and BODY from DXF and DWG

### LINKED ISSUE #935: Support read ACIS data
**Summary**
Support read ACIS data, parser SAB

**Describe**
- I am handling issue #822 but have some problems about read ACIS data.
- When read REGION entities, I had SAB data in ProprietaryData of ModelerGeometry. 
- In the open design doc, AcDb:AcDsPrototype_1b doc seem as only get SAB data in section. May be need parser SAB.
- I don't know how to parser SAB or find document or what is SDK support. Do you have any ideas?

**Additional**

Image

Image

ilCosmico:
Follows up on #935 and #822. This adds the read path for the ACIS payload of the modeler geometry entities across all the container variants, one commit per slice so they can be reviewed on their own.

**API and DXF ACDSDATA section (R2013+)**. `ModelerGeometry.AcisData` (byte[]) carries the raw payload: binary SAB when it starts with the "ACIS BinaryFile" signature, SAT text bytes otherwise. `IsBinaryAcisData` and `GetAcisText()` are the helpers. The new section reader collects the ASM_Data records (owner handle at group 320, hex chunks at group 310) and attaches the blobs by handle. `ProprietaryData` is untouched for compatibility.

**DXF codes 1 and 3 (pre-2013)**. The character swap (0x9F minus the code, above space) is decoded and the chunks are joined with the right semantics: group 1 starts a SAT line, group 3 continues the previous one, so payloads split at the 255 character limit reassemble without spurious breaks. BODY now routes through the modeler geometry reader like REGION and 3DSOLID, before this it lost its codes entirely.

**DWG entity stream**. `readModelerGeometryData` stopped right after the version field. Version 1 (R2000 and older) reads the size-prefixed swapped SAT blocks; version 2 (R2004+) reads the raw ACIS file up to the end marker, which spells "ACIS" or "ASM" depending on the kernel that wrote the payload, in plain ASCII or as a tagged compound name.

**DWG AcDs data section (R2013+)**. The "has DS binary data" flag is now captured and the new AcDb:AcDsPrototype_1b reader parses the segment index and the record directories of the "_data_" segments, which carry the owner handle of every payload, so the association is exact. Payloads sit inline (4 byte size prefix) or in the blob01 segments (4 byte marker plus 8 byte size); the payload area base is anchored on the SAB signatures and each record is validated before it is accepted. Files written since 2018 use the "ASM BinaryFile" signature.

Validated against a collection of 306 real files (R14 to R2018, several producers) and against the sample files already in the repository: more than 16000 payloads extracted, every one of them tokenizes cleanly to its End-of-data marker, and the only entities left empty are the ones whose ACIS empty flag is set in the stream. Known limit: a payload split across multiple blob01 segments is not stitched back (about 0.5% of the collection); the reader notifies the entities left without data.

Read only: writing the payload back and parsing the SAT/SAB content are out of scope, as discussed in #935.

Closes #935

### DIFF
--- a/src/ACadSharp.Tests/IO/AcisTextCodecTests.cs
+++ b/src/ACadSharp.Tests/IO/AcisTextCodecTests.cs
@@ -1,0 +1,82 @@
+using ACadSharp.IO;
+using System.Linq;
+using System.Text;
+using Xunit;
+
+namespace ACadSharp.Tests.IO;
+
+public class AcisTextCodecTests
+{
+	[Fact]
+	public void DecodeIsAnInvolution()
+	{
+		string text = "400 0 1 0\nbody $-1 $1 $-1 $-1 #";
+
+		string swapped = AcisTextCodec.Decode(text);
+
+		Assert.NotEqual(text, swapped);
+		Assert.Equal(text, AcisTextCodec.Decode(swapped));
+	}
+
+	[Fact]
+	public void DecodeBytesMatchesTextDecode()
+	{
+		string text = "body $-1 $1 $-1 $-1 #";
+		byte[] swapped = Encoding.ASCII.GetBytes(AcisTextCodec.Decode(text));
+
+		byte[] decoded = AcisTextCodec.Decode(swapped);
+
+		Assert.Equal(text, Encoding.ASCII.GetString(decoded));
+	}
+
+	[Fact]
+	public void IsEncodedDetectsPlainAndSwappedSat()
+	{
+		string plain = "400 0 1 0";
+
+		Assert.False(AcisTextCodec.IsEncoded(plain));
+		Assert.True(AcisTextCodec.IsEncoded(AcisTextCodec.Decode(plain)));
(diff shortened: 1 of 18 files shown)


ilCosmico:
Hi @DomCR, a heads-up on ordering rather than a review request. The write side of this ACIS work is ready on my end (3DSOLID, REGION and BODY payloads out to DXF and DWG, both the pre-2013 and 2013+ containers), and it builds on this PR plus #1140. I'd rather not stack the write PR on top, so I'm holding it until these two land. I've got a block of time set aside for this right now, so if you can get to #1139 and #1140 soon it would let me rebase onto master and open the write PR while I'm still on it. Happy to make any changes you'd want on these two in the meantime.

DomCR (approved):

DomCR:
Really good job with this one!

PR #1145, opened 2026-07-10
Fix InvalidCastException in ReadHeader() when header variable omits group 30 (Z)

### LINKED ISSUE #1143: DxfReader throws InvalidCastException when header point variable is missing group 30 (Z coordinate)
**Describe the bug**

System.InvalidCastException
   at System.Convert.DefaultToType(IConvertible value, Type targetType, IFormatProvider provider)
   at ACadSharp.Header.CadHeader.SetValue(string systemvar, object[] values) CadHeader.cs line 3344
   at ACadSharp.IO.DxfReader.ReadHeader() DxfReader.cs line 268
   at ACadSharp.IO.DxfReader.Read() DxfReader.cs line 163

DxfReader throws System.InvalidCastException when reading a DXF R12 (AC1003) file whose header point variables — such as $EXTMIN and $EXTMAX — omit the Z coordinate (group code 30). CadHeader.SetValue() attempts to convert the next parsed value to XYZ but receives a Double instead, causing the exception. The entire file is left unread.

**Expected behavior**

The missing group 30 should be treated as 0.0 and the file should parse successfully.

**Additional context**

Encountered with a real-world R12 file produced by a third-party CAD tool. The file omits group 30 for $EXTMIN and $EXTMAX.

A pull request with a fix is in preparation and will be submitted shortly.

dsn27:
# Description

Some DXF files omit group code 30 (Z coordinate) for 2D-only header
variables such as `$EXTMIN` and `$EXTMAX`. When such a variable appears
immediately before `ENDSEC`, `ReadHeader()` reads the `ENDSEC` token (group
code 0, `DxfCode.Start`) as the missing Z parameter and passes the string
`"ENDSEC"` to `CadHeader.SetValue()`. That method calls
`Convert.ChangeType("ENDSEC", typeof(XYZ))`, which throws
`InvalidCastException` and aborts the entire import.

The fix extends the two existing "irregular file" guards in `ReadHeader()` to
also break/skip on `DxfCode.Start` (group 0), not just `DxfCode.CLShapeText`
(group 9). A related type mismatch is fixed at the same time: integer types
(`Int16`/`Int32`/`Int64`) default to `0` (int) while `Double` and `Point3D`
types default to `0d` (double), matching the `XYZ(double, double, double)`
constructor resolved by `SetValue()` via reflection.

# Tasks done in this PR
* [x] An evil bug has been defeated.

- Fixes #1143

# Notes for reviewer
- Only `src/ACadSharp/IO/DxfReader.cs` is changed (3 lines).
- Group code 0 is reserved exclusively for entity/section type identifiers
  by the DXF spec and never appears as a parameter code inside a header
  variable, so extending the guard to `DxfCode.Start` cannot cause early
  exits in well-formed files.
- The `0` vs `0d` split makes the existing `CLShapeText` path more correct
  as well, though that path was not the source of the reported crash.

### DIFF
--- a/src/ACadSharp/IO/DxfReader.cs
+++ b/src/ACadSharp/IO/DxfReader.cs
@@ -222,7 +222,7 @@ namespace ACadSharp.IO
 				{
 					this._reader.ReadNext();
 
-					if (this._reader.DxfCode == DxfCode.CLShapeText)
+					if (this._reader.DxfCode == DxfCode.CLShapeText || this._reader.DxfCode == DxfCode.Start)
 					{
 						//Irregular dxf files may not follow the header type
 						int c = data.DxfCodes[i];
@@ -236,9 +236,11 @@ namespace ACadSharp.IO
 							case GroupCodeValueType.Int16:
 							case GroupCodeValueType.Int32:
 							case GroupCodeValueType.Int64:
+								parameters[i] = 0;
+								break;
 							case GroupCodeValueType.Double:
 							case GroupCodeValueType.Point3D:
-								parameters[i] = 0;
+								parameters[i] = 0d;
 								break;
 							case GroupCodeValueType.None:
 							case GroupCodeValueType.String:
@@ -263,7 +265,7 @@ namespace ACadSharp.IO
 					this.triggerNotification($"Invalid value for header variable {currVar} | {parameters.FirstOrDefault()}", NotificationType.Warning, ex);
 				}
 
-				if (this._reader.DxfCode != DxfCode.CLShapeText)
+				if (this._reader.DxfCode != DxfCode.CLShapeText && this._reader.DxfCode != DxfCode.Start)
 				{
 					this._reader.ReadNext();
 				}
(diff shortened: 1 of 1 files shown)


src/ACadSharp/IO/DxfReader.cs
```diff
@@ -236,9 +236,11 @@ public override CadHeader ReadHeader()
 							case GroupCodeValueType.Int16:
 							case GroupCodeValueType.Int32:
 							case GroupCodeValueType.Int64:
+								parameters[i] = 0;
+								break;
 							case GroupCodeValueType.Double:
 							case GroupCodeValueType.Point3D:
-								parameters[i] = 0;
+								parameters[i] = 0d;
```
DomCR, line 241: The modifications in the switch can be reversed, there is no point to add the int case.

DomCR (changes requested):

DomCR:
The implementation looks good to solve the issue, reverse the switch and is good to go.

Thanks for the contribution!

dsn27:
The separate Int case has been removed.

DomCR (approved):

Issue #1131, opened 2026-07-05
【功能需求】直接解析原生天正 TCH 自定义实体（无需转 T3、无需炸开图纸），支持导出结构化 JSON（[FEATURE] Parse native TArch TCH custom entities from original un-converted DWG (no T3 export / explode) for structured JSON output）


xjtcnj:
在中国，90% 以上的建筑图纸都由天正建筑绘制。目前全球暂无任何工具可以直接解析原生天正 DWG 文件里的 TCH 自定义实体参数，提取完整构件参数并输出结构化 JSON。现有方案只能先将图纸导出为 T3 格式再提取数据，才能生成结构化 JSON。该流程不仅增加额外人工操作，还会不可逆地丢失天正自定义构件的内置参数。如果 ACadSharp 能够支持直接解析原生天正 DWG，将会大幅提升它在国内建筑工程开发从业者中的使用率。举个例子：我本身是建筑师，并不具备编程能力，对此功能需求十分迫切。
More than 90% of architectural drawings in China are created with Tangent TArch. There is currently no tool worldwide that can directly read all complete parameters of TCH custom entities inside original native TArch DWGs and generate structured JSON.
The existing workflow requires converting drawings to T3 format before data extraction. This brings extra manual work and causes permanent, unrecoverable loss of component parameters.
If ACadSharp supports direct parsing of original Tangent DWG files, its popularity among Chinese AEC developers will be greatly improved.
I am an architect without programming background, and this feature is highly needed for my daily work.

DomCR:
Hi @xjtcnj,

I've tried to look for information about the TArch entities but I couldn't find any kind of documentation or file to use as an example.

Could you provide some additional links for more context? 
A file with this sort of entities will help too. I'm guessing that the entities use Proxy graphics which were implemented not long ago so we may be able to read them.

About the Json exportation, I'm working on this feature right now, is not fully integrated yet though.

xjtcnj:
File source: Original DWG saved directly by Tangent TArch (T8/T20/T30 series), no T3 conversion
Target output: Structured JSON with full architectural component parameters
Official software reference: [https://tangent.com.cn/tzgcptxl](https://link.wtturl.cn/?target=https%3A%2F%2Ftangent.com.cn%2Ftzgcptxl&scene=im&aid=497858&lang=zh)
Sample resource support: I am able to supply original native TArch construction drawing DWGs for testing upon request.

[...]

PR #1148, opened 2026-07-10
fix: add AEC_MODI and AEC_SYMB_SECTION as ProxyEntity in readUnlisted…

WallanceLee:
AEC custom objects (TDbSymbModi, TDbSymbSection) from BIM applications like Revit/Tianzheng are stored in DWG with proxy graphics for fallback display in AutoCAD. ACadSharp previously read these as UnknownEntity, causing their proxy graphics to be lost. Route them through readProxyEntity() so the proxy geometry data is preserved in ProxyGeometries for downstream exporters. 

# Description

I don't have a further knownledge about the case branches, and following is from my test data and notification handler.

# Tasks done in this PR
* [ ] Something awesome.

# Notes for reviewer

### DIFF
--- a/src/ACadSharp.Tests/IO/AecProxyFallbackTests.cs
+++ b/src/ACadSharp.Tests/IO/AecProxyFallbackTests.cs
@@ -1,0 +1,54 @@
+using ACadSharp.Classes;
+using ACadSharp.IO.DWG;
+using Xunit;
+
+namespace ACadSharp.Tests.IO
+{
+	public class AecProxyFallbackTests
+	{
+		[Theory]
+		[InlineData("AEC_MODI", "TDbSymbModi", true)]
+		[InlineData("AEC_SYMB_SECTION", "TDbSymbSection", true)]
+		[InlineData("AEC_DBCONFIG", "TDbConfig", true)]
+		[InlineData("AEC_DOOR", "AecDbDoor", true)]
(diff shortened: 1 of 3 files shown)


WallanceLee:
@DomCR Could you review this PR and let me know if you need some modification?

DomCR:
Hi @WallanceLee,

Could you provide a file that contains those entities?

Your solution may work for these entities but given that they are part of AEC, maybe there is a more generic and better solution to include the different objects that haven't been implemented like the `Wall`.

Thanks!

WallanceLee:
[test.dwg.zip](https://github.com/user-attachments/files/30259620/test.dwg.zip)

WallanceLee:
@DomCR @domeitzinger  Hi, I have rewritten all with more general method. Could you give me some suggestions?

WallanceLee:
Any news?

DomCR:
Hi @WallanceLee,

I don't think this is a good approach for the AEC and TArch entities or objects are not written as proxies, they have it's own structure.

As an example you can check the `Wall` AEC entity.

Related issue:
- https://github.com/DomCR/ACadSharp/issues/1131

Also, the current implementation stores the Proxy Graphics when reading the different entities so it should not be a difference if they have stored graphics.

DomCR:
If you want to procede with this branch, you can try to explore the data from each individual component and compare it to the entity in the file, and create their own reading method.

Let me know how you want to continue or if I can help you in any way.

Issue #452, opened 2024-09-17
Drawing file is not valid


amirj0914:
Autocad gives the error after wrriting DWG file with this simple code : 

`

            string dwgFilePath = @"G:/output.dwg";

            CadDocument doc = new CadDocument(ACadVersion.AC1024);

            //Create a point located in (10, 10, 0)
            Point pt = new Point
            {
                Location = new CSMath.XYZ(10, 10, 0)
            };

            //Create a line from the origin to the point (5, 5, 0)
            Line line = new Line
            {
                StartPoint = CSMath.XYZ.Zero,
                EndPoint = new CSMath.XYZ(5, 5, 0)
            };

            doc.Entities.Add(pt);
            doc.Entities.Add(line);

            DwgWriter writer = new DwgWriter(dwgFilePath, doc);
`
After opening the exported file with Autocad 2017 :

![err](https://github.com/user-attachments/assets/affd0305-b5c8-410f-b22b-7bd3cef98c1c)

thaiguer:
Hello, @amirj0914

Make shure you are using the latest version: dotnet add package ACadSharp --version 1.0.0

I created a new file with the following code, and I had success opening it with BricsCad.

`using ACadSharp;
using ACadSharp.Entities;
using ACadSharp.IO;

//old line:
//string dwgFilePath = @"G:/output.dwg";

//new line:
string dwgFilePath = @"C:\dev\output.dwg";

CadDocument doc = new CadDocument(ACadVersion.AC1024);

//Create a point located in (10, 10, 0)
Point pt = new Point
{
    Location = new CSMath.XYZ(10, 10, 0)
};

//Create a line from the origin to the point (5, 5, 0)
Line line = new Line
{
    StartPoint = CSMath.XYZ.Zero,
    EndPoint = new CSMath.XYZ(5, 5, 0)
};

doc.Entities.Add(pt);
doc.Entities.Add(line);

DwgWriter writer = new DwgWriter(dwgFilePath, doc);

//new line
writer.Write();`

There are a few diferences to your code. First, the dwgFilePath, the way you wrote it, seems to be incorrect. And in the end, you forgot to use the method "Write", from the class DwgWriter.

thaiguer:
Hello, @amirj0914

Make shure you are using the latest version: dotnet add package ACadSharp --version 1.0.0

I created a new file with the following code, and I had success opening it with BricsCad.

`Hello, @amirj0914

[...]

PR #1150, opened 2026-07-10
Write the ACIS payload of 3DSOLID, REGION and BODY to DXF and DWG

ilCosmico:
This is the write side of #1139 (reading the ACIS payload of 3DSOLID, REGION and BODY), now merged. The DWG and DXF writers used to skip those entities; they now write the ACIS payload back.

The payload is the same SAT or SAB byte stream regardless of the file type. Only the container changes with the version, so the writer puts the bytes where each version expects them:

- Pre-2013 DXF: SAT text in group codes 1 and 3, with the character swap those codes use.
- Pre-2013 DWG: the payload in the entity stream, SAT text in the version 1 block or the raw stream in the version 2 block.
- 2013 and later DXF: the binary SAB in the ACDSDATA section.
- 2013 and later DWG: the binary SAB in the AcDs data section, which was not written before. The segment directory carries the owner handle, so each payload maps back to its entity, and the payload area is anchored from the directory size so the offsets hold for any record count.

The R2013 and later modeler geometry entity read path is aligned with the same layout (no ACIS empty bit and no silhouette block before the wireframe), so the reader and writer agree.

Tests cover the SAT text codec, the DXF ACDSDATA section, and the DXF SAT round-trip, next to the existing writer tests.

One dependency worth noting: R2013 and later DWG files also need the small viewport reference-type fix in #1149 to open cleanly in AutoCAD.


ilCosmico:
Rebased on current master (which now includes #1149) and added one commit. Following the discussion in #1152, I audited the AcDs writer field by field against chapter 24 of the .dwg file specification, and that audit found one real defect: in the search segment, the per-handle index stored the record's directory position instead of its rank in the sorted handle order. The two coincide when the directory is already sorted by handle, which is why every file opened fine so far; a file saved by AutoCAD with out-of-order handles shows the correct behavior. The new commit fixes it and adds regression tests built from that file.

ilCosmico:
Rebased on current master (3.6.51). Beyond the rebase the only change is in the AcDs round-trip tests: they now read the written section back through the Prototype1b reader, since the old AcDs reader went away with [#1154](https://github.com/DomCR/ACadSharp/pull/1154). The writer itself is untouched, and the build and those tests are green on the new master.

DomCR:
Hi @ilCosmico,

I'm currently looking into the implementation of Prototype1b, I won't be able to check the writer until the reader is stable.

Keep in mind that the writer is always more difficult to implement, there has to be a manual testing with different softwares to ensure that the fil

I'll come back to this one after I'm more familiar with the Prototype1b section, thanks.

ilCosmico:
Hi @DomCR, understood, and I appreciate how careful you're being with the checks. 
I've also run some testing myself already, reading the output files back with DWG TrueView and with an ODA based reader, so there's already a decent amount of validation behind this. 

No rush on my end, I'm on vacation for the next week. 

Thanks for the update!

ilCosmico:
Rebased on today's master. The first three commits are the same work moved onto the reworked writers: `DwgObjectWriter.Common.cs` is gone upstream, so the bit that announces the payload moved into `DwgObjectWriter`, and the entity now passes the `IsValid` gate before reaching the modeler geometry case.

Two commits are new, both found by opening the written files in Autodesk DWG TrueView. A long SAT line is cut at a space between groups 1 and 3, so a token never spans two groups. The `AcDb3dSolid` subclass is written only from the 2007 format: before that its only field does not exist, and the marker on its own makes AutoCAD discard the whole DXF.

Files were checked at R2000, R2004, R2007, R2010 and R2013, in both formats, and they all open. The one file missing is the R2007 DWG, which this writer does not support yet (#452).

PR #1153, opened 2026-07-14
fix binary chunk length for dxf ascii

simonedd:
Split extended-data / binary chunks at 127 bytes instead of 64 in DxfAsciiWriter


DomCR (approved):

DomCR:
Good one! thanks!

PR #1155, opened 2026-07-20
Read the surface entity family as modeler geometry

ilCosmico:
This adds read support for the surface entities: SURFACE, PLANESURFACE, EXTRUDEDSURFACE, LOFTEDSURFACE, REVOLVEDSURFACE, SWEPTSURFACE and NURBSURFACE. Until now the DWG reader turned them into UnknownEntity and the DXF reader discarded them, so their ACIS payload was lost.

The new classes derive from ModelerGeometry, and that is most of the trick: the payload extraction already in place works unchanged. Pre-2013 files carry the payload in the entity itself, 2013+ files in the AcDs data section (DWG) or ACDSDATA (DXF), and both attach points are generic over ModelerGeometry.

In detail:

- `Surface` carries the U/V isoline counts, one class per derived type, mapped by DXF name and subclass marker.
- DWG: the surface types are routed in `readUnlistedType` to a reader that consumes the AcDbModelerGeometry common data. The subclass tails (sweep, loft and revolve parameters) are not decoded yet; the object data is length delimited, so the unread tail is skipped safely.
- DXF: the surface types go through `readEntity` with the modeler geometry template, so codes 1/3 accumulate the SAT text and the subclass codes resolve through the map of the subclass being read.
- Tests: in-memory DXF cases for the seven types, plus two DWG samples with loose extruded sheets (`samples/surfaces`), one saved as 2010 with the payload embedded in the entity and one as 2013 with the payload in the AcDs section.

The write side is out of scope here: today a Brep is always written as 3DSOLID, and the surface subclass fields are not decoded yet. Both are planned as a follow-up, the same read-first sequence used for the ACIS payload extraction.


ilCosmico:
Rebased on current master (3.6.51). Beyond the rebase the only change is in the surface read test: on R2013+ it now takes the ACIS payload from `DataStorage` by handle, since [#1154](https://github.com/DomCR/ACadSharp/pull/1154) no longer populates `AcisData` there. The reader code is untouched, and the build and the surface tests are green on the new master.

ilCosmico:
Hi @DomCR, a gentle nudge on this one whenever you have a moment. It is additive: the surface entity classes plus two hooks in the readers. It does not touch the Prototype1b or ACDS code you are reworking in #1174, so it should stay out of the way of that refactor, and CI is green with the branch still merging cleanly.

On my side this is the last piece missing to read the surface family, so it would close the reading work.

ilCosmico:
Hi @DomCR, just a line to say that this one still merges cleanly on today's master, after all the movement of the last two weeks, so looking at it should not cost you any rebase work. If you would rather have it split into smaller pieces or reshaped in some way, tell me and I will do that.

PR #1156, opened 2026-07-20
Read the AcDs data blob reference records from the blob01 pages

ilCosmico:
In the AcDs data section a record whose payload does not fit in the data segment is stored as a data blob reference: the size field carries the fixed marker `0xBB106BB1`, followed by the total size and the list of pages, and each page lives in a `blob01` file segment addressed through the segment index. The reader did not know the marker, validated the record as a plain inline payload, failed, and dropped it. The owning entity came back with no ACIS payload.

I hit this on a 2013 file where one solid out of 89 carried a ~407 KB payload: that one entity was skipped while everything else loaded fine.

The fix keeps the segment index offsets around while reading the directories, recognizes the marker during record validation, and assembles the payload from the blob01 pages. Each page declares its own start offset, so the stream order does not matter. A synthetic two-page test covers the reassembly.

### DIFF
--- a/src/ACadSharp.Tests/IO/DWG/DwgAcDsDataReaderTests.cs
+++ b/src/ACadSharp.Tests/IO/DWG/DwgAcDsDataReaderTests.cs
@@ -70,6 +70,87 @@ public class DwgAcDsDataReaderTests
 		Assert.Empty(DwgAcDsDataReader.ReadRecords(buffer));
 	}
 
+	[Fact]
+	public void ReadRecordsAssemblesBlobReferencePayloadFromPages()
+	{
+		// A payload too large for the data segment becomes a data blob
+		// reference record (its size field carries the 0xBB106BB1 marker) whose
+		// pages live in blob01 segments addressed through the segment index.
+		// Two pages, out of stream order, check the reassembly.
+		byte[] acis = Encoding.ASCII.GetBytes("ACIS BinaryFile-blob-payload-split-into-two-pages");
+		int firstPageSize = 20;
+		int secondPageSize = acis.Length - firstPageSize;
+
+		using MemoryStream stream = new MemoryStream();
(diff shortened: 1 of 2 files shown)


DomCR:
Hi @ilCosmico,

This PR collides with:
- https://github.com/DomCR/ACadSharp/pull/1154

which I'm reviewing for the Prototype1b implementation. Do we need this one if the other gets approved?

ilCosmico:
Thanks @DomCR. Agreed that #1154 is the better base for the section itself: parsing the whole Prototype1b into DataStorage, with the blob01 pages reassembled, covers more than the incremental reader.

One catch before we close #1156, though. #1154 also drops the per-entity wiring that #1139/#1146 added.
In DwgReader the `foreach (ModelerGeometry entity ...) entity.AcisData = payload` loop is replaced by just storing `document.DataStorage`, so the parsed ACIS payload ends up on the document but never reaches `ModelerGeometry.AcisData`, which is where downstream code reads the SAB stream for 3DSOLID/REGION/BODY and the surface entities on 2013+ files. It's silent rather than a crash, so a load test wouldn't catch it.

So the two are really complementary: #1154 parses the section, but something still has to hand the ACIS blobs from DataStorage back to `ModelerGeometry.AcisData`. Happy to close #1156 once that's back in, or I can rebase it to do exactly that on top of your DataStorage instead of the old reader. Whichever you'd rather.

#1155 is separate (different files), so it's unaffected either way.

ilCosmico:
Closing as promised. [#1154](https://github.com/DomCR/ACadSharp/pull/1154) is merged and I've now validated the whole path end to end on my side: on R2013+ files the ACIS payloads come back through `DataStorage` exactly as they did before, across the real models in my corpus, so `ResolveDataEntry` fully covers the blob01 reassembly this PR was adding. Thanks @domeitzinger for the accessor.

PR #1162, opened 2026-07-27
read MEASUREMENT from the AcDb:Template section

simonedd:
# Description

Implement readTemplate() (Int16 description length + optional description bytes + UInt16 MEASUREMENT).


simonedd:
Hi @DomCR,

Have you already had a chance to look at this PR? 
In my opinion the change is small and self-contained, and it would unblock an important piece of work I'm doing on hatches (I rely on MEASUREMENT to pick the right predefined pattern).

I also have a few other small, low-impact PRs open. 
Thanks for your work on ACadSharp.

PR #1174, opened 2026-08-02
Refactor ACDS/ACIS schema handling and storage

DomCR:
# Description

Introduce CadFileDataStorage with new Schema and SchemaProperty types for structured ACDS data.
Update DxfAcdsDataSectionReader and document builders to use the new model.

### DIFF
--- a/src/ACadSharp/CadDocument.cs
+++ b/src/ACadSharp/CadDocument.cs
@@ -209,7 +209,7 @@ public class CadDocument : IHandledCadObject
 	/// <summary>
 	/// The data stored in the Prototype1b header section. This primarily contains ACIS and thumbnail data
 	/// </summary>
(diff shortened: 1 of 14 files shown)


ilCosmico:
Hi @DomCR, I checked out this branch and ran the tests around the ACDS area on net9.0: 10 pass, and `DxfAcdsDataSectionTests.ReadAcdsDataAttachesAcisPayloadToOwnerEntity` fails with `Assert.NotNull() Failure: Value is null`, since `readAcdsRecord` throws `NotImplementedException`. Compared to master the DXF side loses the payload on files saved from 2013 onwards.

That path came in with #1139, so I know it well and I am glad to port it onto the new schema types if that helps.
If you would rather carry the refactor on your own way and at your own pace, that is fine too, just tell me and I will leave it with you.

PR #1192, opened 2026-08-18
A table with more than 32767 entries can be written to DXF

redbluevn:
# Description

Group code 70 of a table header is a 16 bit value, and `DxfTablesSectionWriter` passed the entry count to a checked conversion. A real drawing of 17 MB holds 40843 block records, so writing it to DXF threw `OverflowException: Value was either too large or too small for an Int16` and no file came out at all.

# Tasks done in this PR

* [x] Let the count wrap round, which is what AutoCAD does.

- None; this one stands on its own and is cut straight from `master`.

# Notes for reviewer

**Testing**

`dotnet test` on this branch: **2312 passed / 17 failed**, the same as master.

Oracle: asked to save the same drawing as DXF, AutoCAD writes **-24493** for its 41043 block records. The number is advisory - the entries that follow the header are what a reader counts.

The drawing that could not be written now produces a 190 MB DXF in 17 seconds with no writer warnings, and its entity count survives the round trip with no loss inside the 2D scope.

- Found while running the library over real production drawings rather than the sample set.

---

_Checked against AutoCAD 2027 through `accoreconsole`: AUDIT for "does it open cleanly", and SAVEAS DXF for the oracle comparisons quoted above._

### DIFF
--- a/src/ACadSharp/IO/DXF/DxfStreamWriter/DxfTablesSectionWriter.cs
+++ b/src/ACadSharp/IO/DXF/DxfStreamWriter/DxfTablesSectionWriter.cs
@@ -37,7 +37,7 @@ namespace ACadSharp.IO.DXF
 
 			this._writer.Write(DxfCode.Subclass, DxfSubclassMarker.Table);
 
-			this._writer.Write(70, table.Count);
+			this._writer.Write(70, table.Count > short.MaxValue ? (short)0 : (short)table.Count);
 
 			if (!string.IsNullOrEmpty(subclass))
 			{


src/ACadSharp/IO/DXF/DxfStreamWriter/DxfTablesSectionWriter.cs
```diff
@@ -37,7 +37,12 @@ private void writeTable<T>(Table<T> table, string subclass = null, bool writeFla
 
 			this._writer.Write(DxfCode.Subclass, DxfSubclassMarker.Table);
 
-			this._writer.Write(70, table.Count);
+			//Group code 70 of a table header is a 16 bit value, and a real drawing can hold more
+			//entries than that: an interior drawing of 17 MB in the corpus has 40843 block records.
+			//Converting that with a checked conversion threw and no DXF came out at all. AutoCAD
+			//lets the count wrap round instead - it writes -24493 for its 41043 block records - and
+			//the number is advisory anyway, the entries that follow are what counts.
+			this._writer.Write(70, unchecked((short)table.Count));
```
DomCR: this writes a negative value, which is not correct, the behavior that dxf has when the value excedes the short max value is to write a 0, it would be best to check for the value boundary rather than write a wrong value.

Please remove the comment too, is not needed here, thanks.

redbluevn:
Done, both points: the count is now `table.Count > short.MaxValue ? (short)0 : (short)table.Count`, and the comment is gone.

One measurement for the record, since it surprised me and may be useful to you: asked to save the same drawing as DXF, AutoCAD 2027 writes the wrapped value itself — `70 = -24493` for its 41,043 block records — which is why the original patch matched it. I checked both forms against AutoCAD on that drawing (40,843 block records here) and it opens the DXF with 0 audit errors either way, so nothing depends on the choice and writing 0 is the safer of the two to publish. Happy with it as it stands.

DomCR (approved):

DomCR:
Merging this one, I'll be taking a look at the others when I can.

Thanks for the contribution!

PR #1197, opened 2026-08-19
A model space named in upper case still writes its entities to the entities section

redbluevn:
# Description

The blocks writer decides where a block record's entities belong by comparing its name with
`*Model_Space` and `*Paper_Space`:

```csharp
if (b.Name == BlockRecord.ModelSpaceName || b.Name == BlockRecord.PaperSpaceName)
```

A drawing does not have to spell it that way. AutoCAD matches the name without case, and a
production drawing here carries `*MODEL_SPACE` in upper case - the reader handles it correctly, so
`doc.ModelSpace` is that record and holds all 3,328 entities. The writer did not, so every one of
them went into the BLOCKS section as an ordinary block, and the ENTITIES section came out with a
single record.

AutoCAD reads that back and draws it - the round trip audits 0 errors and plots within 0.4% of the
source - so nothing looks wrong from inside. Any other consumer that reads the ENTITIES section,
which is where the format puts model space, sees an empty drawing.

# Tasks done in this PR

* [x] The model space and paper space names are matched without case.
* [x] `DxfModelSpaceNameTests`: a drawing whose model space is `*MODEL_SPACE` writes its line to the
      ENTITIES section, the usual spelling is unchanged, and an ordinary block still keeps its
      entities in the BLOCKS section.

- None.

# Notes for reviewer

**Testing**

`dotnet test` on this branch: **2315 passed / 17 failed**, the same 17 failures as master
(2312 + 3 new tests).

Measured on the production drawing that carries the upper-case name: the DXF round trip goes from
**1 record in the ENTITIES section to 3,407** - the same count AutoCAD writes for the same drawing
in its own DXF. AUDIT stays at 0 errors before and after, on that drawing and on the sample set.


src/ACadSharp/IO/DXF/DxfStreamWriter/DxfBlocksSectionWriter.cs
```diff
@@ -54,7 +55,12 @@ private void writeBlock(Block block)
 
 		private void processEntities(BlockRecord b)
 		{
-			if (b.Name == BlockRecord.ModelSpaceName || b.Name == BlockRecord.PaperSpaceName)
+			//The name is matched without case: AutoCAD does, and real drawings carry *MODEL_SPACE
+			//in upper case. Comparing exactly wrote every model entity of such a drawing into the
+			//BLOCKS section as an ordinary block and left the ENTITIES section empty - AutoCAD
+			//still draws it, but a consumer that reads the entities section sees nothing.
```
DomCR: The comment is not needed.

src/ACadSharp.Tests/IO/DXF/DxfModelSpaceNameTests.cs
```diff
@@ -0,0 +1,115 @@
...
+namespace ACadSharp.Tests.IO.DXF
+{
+	/// <summary>
+	/// A drawing does not have to spell its model space block "*Model_Space": AutoCAD matches the
+	/// name without case, and real drawings carry "*MODEL_SPACE". The DXF writer compared the name
+	/// exactly, so for such a drawing every model entity was written inside the BLOCKS section as
+	/// an ordinary block and the ENTITIES section came out empty - AutoCAD still draws it, but any
+	/// consumer that reads the ENTITIES section sees an empty drawing.
+	/// </summary>
+	public class DxfModelSpaceNameTests
```
DomCR: Remove these tests.

For these particular cases I usually have a full file in my local tests, there is no need for a specific test in master.

DomCR (approved):

src/ACadSharp/IO/DXF/DxfStreamWriter/DxfBlocksSectionWriter.cs
redbluevn: Removed.

src/ACadSharp.Tests/IO/DXF/DxfModelSpaceNameTests.cs
redbluevn: Removed the file. Suite: 2312 passed / 17 failed - the same failures as master, none added.

PR #1203, opened 2026-08-23
Issue 278 student version ac1032

### LINKED ISSUE #278: DwgWriter using  ACadVersion show his drawing was saved with a student Version
Hi

If I use ACadVersion then I get = his drawing was saved with a student Version
but it is ok if I don't use ACadVersion

`    

    Const dwgFilePath22 = "Test1.dwg"
    doc.Header.Version = ACadVersion.AC1032

    Using writer As DwgWriter = New DwgWriter(dwgFilePath22, doc)
        writer.Write()
    End Using
End Sub`

Regards
Sigster

### LINKED ISSUE #1169: User has reported an exported DWG is saved with 'Student Version of AutoDesk'
This is an odd one, and could well be written off as some bizarre 'user error' in terms of what they did with the DWG after we created it.

But I've had someone report one of the DWG's created via ACadSharp showed this prompt when they opened it (I'm verifying what tool they used to open it)

Image

Has anyone seen this before? Is there something in the header of the CAD document we need to set?

DomCR:
# Description

Fix for the student version flag for AC1032.

### DIFF
--- a/src/ACadSharp/ACadSharp.csproj
+++ b/src/ACadSharp/ACadSharp.csproj
@@ -17,7 +17,7 @@
 	<PropertyGroup>
 		<GenerateDocumentationFile>true</GenerateDocumentationFile>
 		<PackageReadmeFile>README.md</PackageReadmeFile>
-		<Version>3.7.6</Version>
+		<Version>3.7.7</Version>
 		<PackageOutputPath>../nupkg</PackageOutputPath>
 		<SignAssembly>True</SignAssembly>
 		<AssemblyOriginatorKeyFile>../ACadSharp.snk</AssemblyOriginatorKeyFile>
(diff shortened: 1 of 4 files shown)


DomCR:
@IainS1986, could you test this branch and check if it solves the issue for the student version? thanks

IainS1986:
Hey, I can try - not 100% when I'll get a few moments.

Only issue is the only way I could verify the issue (it got reported by a user, I never actually saw the error) was to use a free trial of AutoCAD which has now expired 😂

I'll see if I can setup another trial or find someone else to do one to verify.

Even then AutoCAD didn't show any errors, I was only able to verify it apparently happening using the command `STUDENTDRAWING` in AutoCAD and seeing one export version had `1` and one had `0` 🤡

DomCR:
I didn't know about `STUDENTDRAWING` with this I should be able to check the result, thanks!

PR #1204, opened 2026-08-24
Add DWG reading for Mechanical BOM objects

olix-cn:
# Description

Adds DWG read support for AutoCAD Mechanical BOM data. It registers and reads `ACMBOM`, `ACMBOMROW`, `ACMDATAENTRYBLOCK`, and `ACMDATAENTRYPART`, then resolves their relationships with part references, balloons, and part lists.

This follows the Mechanical entity support introduced in #1123 and exposes structured BOM data instead of leaving these objects unread.

# Tasks done in this PR

* [x] Added public object models and DXF class metadata for Mechanical BOM and data-entry objects.
* [x] Added DWG reader templates and relationship resolution for BOM rows, part references, balloons, and part lists.
* [x] Made Mechanical entity handle parsing consume the complete handle section.
* [x] Added model and metadata tests for the new objects.

- Follow-up to #1123.

# Notes for reviewer

- The decoded layout was validated with a private AC1032 AutoCAD Mechanical drawing containing one BOM and three rows. The fixture is not included because it contains proprietary drawing data.
- Integer fields that are not documented by the public Mechanical format are retained in `RawValues` and `RawAttributeValues` for forward analysis.
- The new `AcmBomTests` pass on both `net9.0` and `net48` (3/3 on each target).
- Local full-suite runs produced the same result on both targets: 2319 passed, 2 skipped, and 17 failures in existing DXF dictionary and MultiLeader round-trip cases outside the Mechanical BOM paths.

### DIFF
--- a/src/ACadSharp.Tests/Objects/Mechanical/AcmBomTests.cs
+++ b/src/ACadSharp.Tests/Objects/Mechanical/AcmBomTests.cs
@@ -1,0 +1,82 @@
+using ACadSharp.Objects.Mechanical;
+using Xunit;
+
+namespace ACadSharp.Tests.Objects.Mechanical;
+
+public class AcmBomTests
+{
+	[Fact]
+	public void MechanicalBomObjectsHaveExpectedClassNames()
+	{
+		AssertClassNames(new AcmBom(), DxfFileToken.AcmBom, DxfSubclassMarker.Bom);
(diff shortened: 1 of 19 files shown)


DomCR:
Hi @olix-cn,

Could you share the file that you use to test this feature?

Thanks!

PR #1206, opened 2026-08-25
Dynamic block data is written by default, and the DXF writer honours the switch

redbluevn:
﻿
# Description

`CadWriterConfiguration.WriteDynamicBlockData` defaults to `false`, and the DWG writer obeys it: the
`EvaluationGraph`, every `BlockRepresentationData` and the `DynamicBlockPurgePreventer` are left out.
A document read from a DWG and written back to a DWG with no configuration at all therefore loses
**every dynamic block it has** â€” the block comes back static, without its parameters, actions and
grips. `samples/dynamic-blocks/BLOCKFLIPPARAMETER.dwg` round-trips to `EvaluationGraph == null`.

The DXF writer had the same three cases in its `isObjectSupported` switch, but they shared the
`default` branch: the data was written whatever the configuration said. That is why a DXF kept the
dynamic blocks the DWG lost, from the same document and the same settings.

The remark on the property says the default is off because the graph *"can cause corruption for
some documents"*. Measured with AutoCAD 2027 (`AUDIT` through `accoreconsole`), writing it on:

| files written with the data on | AUDIT |
|---|---|
| the 10 `samples/dynamic-blocks/*.dwg`, round-tripped through DWG | 10 Ã— **0 errors**; AutoCAD's own DXF save of each still carries `ACAD_ENHANCEDBLOCK` and the parameter/action objects â€” the block is still dynamic |
| 17 production drawings (2 to 25 dynamic blocks each; 132 in all), 51 channels: DWG R2018, DWG R2000, DXF R2018 | **every channel audits to the same number it audited to with the data off** â€” not one error more |

| where | after |
|---|---|
| `CadWriterConfiguration.WriteDynamicBlockData` | default `true`; remark rewritten with the measurement |
| `DxfObjectsSectionWriter.isObjectSupported` | the three cases return `false` when the switch is off, as the DWG writer does |

Switching it off gives the old DWG behaviour back for a document that turns out to need it â€” and
now gives the same in DXF.

# Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

The default changes; a caller that left it alone now gets the dynamic blocks it was silently losing.
A caller that set it explicitly sees no change in DWG and, in DXF, finally gets what the setting says.

# How Has This Been Tested?

`DynamicBlockRoundTripTests` (new): for each of the ten dynamic-block samples the graph read back
from a DWG and from a DXF written with the default configuration has the same nodes (id, flags,
expression type, order) and the same edges as the one read from the sample; with the switch off
both writers leave the graph out; the default is `true` for both configurations. The DWG half of
`BlockActionReferencesTests`, which had to be left out while the DWG lost the graph, is added.
Whole suite: see the checklist table. AutoCAD measurements above.

# Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective
- [x] New and existing unit tests pass locally with my changes


src/ACadSharp/IO/DXF/DxfStreamWriter/DxfObjectsSectionWriter.cs
```diff
@@ -725,6 +719,13 @@ private bool isObjectSupported(CadObject co)
 		{
 			case UnknownNonGraphicalObject:
 				return false;
+			//The switch is honoured here as it is in the DWG writer. Before, these three cases shared
+			//the default branch and were written whatever the configuration said - so the DXF kept
+			//the dynamic blocks that the DWG, with the same document and the same settings, dropped.
+			case EvaluationGraph when !this.Configuration.WriteDynamicBlockData:
+			case BlockRepresentationData when !this.Configuration.WriteDynamicBlockData:
+			case DynamicBlockPurgePreventer when !this.Configuration.WriteDynamicBlockData:
+				return false;
```
DomCR, line 724: The comment can be removed.

The case can be group with the `UnknownNonGraphicalObject` like the others below.

src/ACadSharp/IO/CadWriterConfiguration.cs
```diff
@@ -49,15 +49,22 @@ public class CadWriterConfiguration
...
+	/// ten dynamic-block samples and all seventeen production drawings audit in AutoCAD exactly as
+	/// they did with it off, and AutoCAD keeps the blocks dynamic. Switch it off to get the old
+	/// behaviour back for a document that turns out to need it.
 	/// </remarks>
 	/// <value>
-	/// default: false
+	/// default: true
 	/// </value>
-	public bool WriteDynamicBlockData { get; set; } = false;
+	public bool WriteDynamicBlockData { get; set; } = true;
```
DomCR: This change effects dwg too and the default behavior of the library, please revert all the changes in this file.

src/ACadSharp.Tests/IO/DynamicBlockRoundTripTests.cs
```diff
@@ -0,0 +1,136 @@
...
+
+//The rewrite test in DynamicBlockTests writes every dynamic-block sample with the dynamic data
+//switched on and stops there: nothing reads the file back. With the old default (off) a DWG written
+//by a caller who set nothing came back with no evaluation graph at all, and every dynamic block in
+//it was static from then on. These tests close that loop for both formats:
+//the graph that comes out has the same nodes, in the same order, with the same expression types
+//and flags, and the same edges, as the one that went in. AutoCAD audits every one of the ten
+//written DWGs with 0 errors and still treats the block as dynamic (its own DXF save of the file
+//keeps the ACAD_ENHANCEDBLOCK entry and the parameter and action objects).
+public class DynamicBlockRoundTripTests : IOTestsBase
```
DomCR: There is no need for these tests, please remove this file.

DomCR (approved):

src/ACadSharp/IO/DXF/DxfStreamWriter/DxfObjectsSectionWriter.cs
redbluevn: Done - grouped with UnknownNonGraphicalObject and the comment removed.

src/ACadSharp/IO/CadWriterConfiguration.cs
redbluevn: Reverted, the default stays off. One note for the record: the old guarded cases fell through to the default branch either way, so the DXF writer used to write these three objects whatever the flag said; with the grouping above it now honours the flag. If you would rather the DXF writer keep the old always-write behaviour, say the word and I will drop that part too.

src/ACadSharp.Tests/IO/DynamicBlockRoundTripTests.cs
redbluevn: Removed. Suite on the branch: 2313 passed / 17 failed - the same failures as master, none added.

PR #1215, opened 2026-08-25
A data store property with more than one value keeps all of them

redbluevn:
﻿
> Root branch, cut from upstream master. Independent of the AcDs writer chain.

# Description

`DwgPrototype1bReader.readSchemaProperty` sizes the value array by the count it just read, then
writes **every** value it reads into row zero:

```csharp
prop.Values = new byte[prop.PropertyValueCount, prop.TypeSize];
for (int i = 0; i < prop.PropertyValueCount; i++) {
    byte[] propertyValue = this._reader.ReadBytes((int)prop.TypeSize);
    for (int j = 0; j < propertyValue.Length; j++) {
        prop.Values[0,j] = propertyValue[j];   // always row 0
    }
}
```

So a property with two values comes back with the second value sitting in the first row and the
second row empty.

This is the common case rather than a corner of one: **every schema in a data store carries an
`AcDbDs::ID` property and it always has two values**. Read back, a store looks as though its schemas
were numbered from zero, when the number is in the value the row overwrote.

# Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Suite **2314 passed / 17 failed** on upstream `edda5594`, whose own baseline is 2313 / 17 - one case
added, none broken.

The new test reads `sample_AC1032.dwg` and asks that every row the value count promised holds
something. It **fails before the change and passes after it**.

# Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective
- [x] New and existing unit tests pass locally with my changes


src/ACadSharp.Tests/IO/AcDsSchemaPropertyTests.cs
```diff
@@ -0,0 +1,41 @@
...
+using Xunit;
+
+namespace ACadSharp.Tests.IO;
+
+/// <summary>
+/// Every schema in a data store carries an AcDbDs::ID property, and it always has two values.
+/// They were read into the same row of the array the value count had just sized, so the first
+/// came back overwritten by the second and the second row came back empty.
+/// </summary>
+public class AcDsSchemaPropertyTests
```
DomCR: You can remove the tests this part of the code will have a major refactor.

redbluevn: Done in 4d2f8944 — test removed. What is left in the PR is the one-line reader fix in `DwgPrototype1bReader`.

PR #1220, opened 2026-08-25
Writing a DXF as AC1014 says so instead of handing back a file AutoCAD refuses

redbluevn:
﻿
# Description

`DwgWriter` already tells the caller that an `AC1014` file is not accepted by AutoCAD. `DxfWriter`
wrote the same version in silence, and the result is refused just as firmly. AutoCAD 2027 on a DXF
we wrote as AC1014:

```
Premature end of object on line 2906.        (the DIMSTYLE table header)
Invalid or incomplete DXF input -- drawing discarded.
```

Removing the group it stops on only moves the refusal to the next object (`ACDBPLACEHOLDER`
wanting its class separator), so the R14 DXF shape is wrong in more than one place.

Making an R14 DXF that opens is separate work, and there is no oracle for it: AutoCAD 2027 cannot
save that version either, so there is nothing to compare against. What this fixes is the silence -
a caller asking for a version whose output cannot be opened is now told, exactly as on the DWG side.

# Tasks done in this PR

* [x] `DxfWriter.Write` reports `AC1014` and older as not accepted by AutoCAD, naming `AC1015` as
      the oldest version that opens.
* [x] `DxfVersionWarningTests`: AC1014 is reported; AC1015, AC1018 and AC1032 are not.

- None.

# Notes for reviewer

**Testing**

`dotnet test` on this branch: **2316 passed / 17 failed**, the same 17 failures as master
(2312 + 4 new tests).

How it was found: one drawing written to every version, each result audited by AutoCAD.

| version | DWG | DXF |
|---|---|---|
| AC1014 | not opened (already warned) | **not opened, no warning** |
| AC1015, AC1018, AC1024, AC1027, AC1032 | opened, 0 errors | opened, 0 errors |
| AC1021 | write throws `CadNotSupportedException` | opened, 0 errors |

### DIFF
--- a/src/ACadSharp.Tests/IO/DXF/DxfVersionWarningTests.cs
+++ b/src/ACadSharp.Tests/IO/DXF/DxfVersionWarningTests.cs
@@ -1,0 +1,57 @@
+using ACadSharp.Entities;
+using ACadSharp.IO;
+using CSMath;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Xunit;
+
+namespace ACadSharp.Tests.IO.DXF
+{
+	/// <summary>
+	/// The DWG writer says plainly that an AC1014 file is not accepted by AutoCAD. The DXF writer
+	/// wrote the same version in silence, and the result is refused just as firmly: AutoCAD 2027
+	/// reports "Premature end of object" on the DIMSTYLE table header, and removing that group only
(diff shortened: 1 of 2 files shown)


DomCR:
Other softwares or older Autocad version accept the version AC1014, this PR is not needed.

PR #1237, opened 2026-09-02
Preserve full-sweep HATCH boundary angles

mediummandoo:
# Description

Preserve full-sweep HATCH arc and ellipse boundaries while keeping DXF angles in `-360..360`. Converting negative angles to positive equivalents can collapse a sweep such as `-90..270` to equal endpoints.

Both edge types use `MathHelper.RadToDeg(..., normalize: true, absolute: false)`. A HATCH-only adjustment preserves full sweeps whose normalized endpoints would otherwise coincide: `180..540` is written as `-180..180`, not `180..180` or an angle above 360. No common utility or other entity conversions are changed.

# Tasks done in this PR

- [x] Move the file-generation fixture into `WriterSingleObjectTests.SingleCaseGenerator` and register it in `Data`.
- [x] Cover full, partial, and zero sweeps, signed angle limits, and direction flags in `DxfWriterSingleObjectTests`.
- [x] Include the negative-angle arc and above-range full ellipse in the shared generation fixture.

- #1252
- DomCR/CSUtilities#26

# Notes for reviewer

`net9.0` writer tests: **859/859 passed** with the repository's CI settings:

```sh
LOCAL_ENV=false SAVE_OUTPUT_IN_STREAM=true SAVE_PREVIEW=false SELF_CHECK_OUTPUT=false \
dotnet test src/ACadSharp.Tests/ACadSharp.Tests.csproj --framework net9.0 \
  --filter 'FullyQualifiedName~DxfWriterSingleObjectTests|FullyQualifiedName~DwgWriterSingleObjectTests'
```

The 12 angle cases assert round-trip results independently of `SELF_CHECK_OUTPUT`. The shared fixture also passes all 11 DXF/DWG file-generation and re-read cases with self-check enabled. `net48` builds successfully. AutoCAD display behavior has not been checked locally.


DomCR:
The methods that you are replacing do the same as the change, it seems a redundant change.

mediummandoo:
Thanks for reviewing. I checked the exact CSUtilities revision pinned by ACadSharp (`fdf1403ede6e3376a0baa1deac166cbe4d21262e`). [`MathHelper.RadToDeg`](https://github.com/DomCR/CSUtilities/blob/fdf1403ede6e3376a0baa1deac166cbe4d21262e/CSMath/MathHelper.cs#L324-L328) is not equivalent to direct multiplication: it always calls `NormalizeAngle(result)`, and its `absolute` parameter is currently unused.

For this full sweep:
- `RadToDeg(-π/2)` produces `270`
- `RadToDeg(3π/2)` also produces `270`
- direct conversion produces `-90` and `270`, preserving the 360-degree span

The [DXF reader](https://github.com/DomCR/ACadSharp/blob/91ae384a4fad23c5a282a47c79c99bdebd4a6b97/src/ACadSharp/IO/DXF/DxfStreamReader/DxfSectionReaderBase.cs#L2138-L2143) only converts those values back with `DegToRad`, so it cannot recover the lost sweep. This is the case covered by the added arc and ellipse regression test.

Would you consider reopening this PR? If you prefer, I can instead fix the unused `absolute` behavior in CSUtilities and update ACadSharp to use `RadToDeg(value, false)`.

DomCR:
I see, the issue then is in CSUtilities, it should allow the option for non absolute values in range of -360 to 360, reopening the PR.

DomCR:
Hi @mediummandoo,

CSUtilities has been updated.

src/ACadSharp.Tests/IO/DXF/DxfHatchWriterTests.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.IO;
+using System.Linq;
+using Xunit;
+
+namespace ACadSharp.Tests.IO.DXF;
+
+public class DxfHatchWriterTests
+{
+	[Fact]
+	public void WriteHatchPreservesFullArcAndEllipseSweeps()
```
DomCR: Please move the file generation test into `WriterSingleObjectTests`.

mediummandoo: Thanks! Moved the generation fixture into `WriterSingleObjectTests` and kept the DXF round-trip assertions in `DxfWriterSingleObjectTests`.

The four HATCH angle conversions now use `RadToDeg(..., normalize: false)` from the updated CSUtilities. Updated in 34bd66b1; all CI checks pass.

src/ACadSharp/IO/DXF/DxfStreamWriter/DxfSectionWriterBase.Entities.cs
```diff
@@ -528,16 +528,16 @@ private void writeHatchBoundaryPathEdge(Hatch.BoundaryPath.Edge edge)
 			case Hatch.BoundaryPath.Arc arc:
 				this._writer.Write(10, arc.Center);
 				this._writer.Write(40, arc.Radius);
-				this._writer.Write(50, MathHelper.RadToDeg(arc.StartAngle));
-				this._writer.Write(51, MathHelper.RadToDeg(arc.EndAngle));
+				this._writer.Write(50, MathHelper.RadToDeg(arc.StartAngle, normalize: false));
+				this._writer.Write(51, MathHelper.RadToDeg(arc.EndAngle, normalize: false));
```
DomCR: To have the angle in range of -360 to 360 the flag `normalize` should be set to `true` and the flag `absolute` to `false`, otherwise for values higher than 360 Autocad will draw a full circle and it will only update the drawing when the arc is modified.

DomCR (approved):

DomCR:
One last change and is good to go!

src/ACadSharp/IO/DXF/DxfStreamWriter/DxfSectionWriterBase.Entities.cs
mediummandoo: Thanks for clarifying — updated in f0bc83aa. Both edge types now use `normalize: true, absolute: false`.

I kept a small HATCH-only full-sweep adjustment so the existing `180..540` case writes `-180..180` rather than equal endpoints. The shared fixture covers this case, and the round-trip tests also check partial arcs, signed limits, and direction flags.

The writer tests and CI pass. I haven't verified AutoCAD display behavior locally.

PR #1241, opened 2026-09-03
fix: correct face-list index advance in proxy shell reader

lchojnack:
# Description

The loop advanced the counter by `count + 1` *in addition to* the `for` loop's own `i++`, so every face over-counted by one. As a result the loop exited before the whole face list was consumed, leaving unread bytes in the stream. Edge/face traits and every subsequent proxy geometry in the same blob were then parsed from the wrong offset, producing garbage entities or read exceptions.

# Tasks done in this PR
* [ ] Something awesome.

# Notes for reviewer


lchojnack:
Hi @DomCR, Please look at this PR. It is quite simple fix. I hope is usefull.

DomCR (approved):

DomCR:
Hi @lchojnack,

I've noticed that the local files that I use for testing do not hit this part of the code, could you provide a file to test this one?

Thanks!

lchojnack:
@DomCR 
When the package with the fix can be released?

PR #1243, opened 2026-09-04
Do not throw on duplicated table entries when Failsafe is enabled

### LINKED ISSUE #1239: DXF read throws when a table has a duplicated entry name and no default entry
Reading a DXF whose LAYER table carries two entries with the same name and no layer "0" throws `KeyNotFoundException: The given key '0' was not present in the dictionary`, and nothing of the file is read. Same file on 3.7.1: read fine.

Minimal file that reproduces it:

```
0
SECTION
2
TABLES
0
TABLE
2
LAYER
70
2
0
LAYER
2
A
70
0
62
7
6
CONTINUOUS
0
LAYER
2
A
70
0
62
7
6
CONTINUOUS
0
ENDTAB
0
ENDSEC
0
SECTION
2
ENTITIES
0
ENDSEC
0
EOF
```

```csharp
using DxfReader reader = new DxfReader(path);
CadDocument doc = reader.Read();   // throws on master, works on 3.7.1
```

Stack:

```
ACadSharp.Tables.Collections.LayersTable.GetDefaultEntry() in LayersTable.cs:24
ACadSharp.Tables.Collections.Table`1.Remove(String key) in Table.cs:137
ACadSharp.IO.DXF.DxfStreamReader.DxfTablesSectionReader.readEntries[T](CadTableTemplate`1)
ACadSharp.IO.DXF.DxfStreamReader.DxfTablesSectionReader.readTable() in DxfTablesSectionReader.cs:128
ACadSharp.IO.DxfReader.Read() in DxfReader.cs:162
```

The duplicate is handled by `Table<T>.Remove`, which calls `assignToDefault`, and that asks for the default entry while the table is still being read, before the default is created. Two neighbouring cases read without a problem, which is what points at the ordering:

- duplicated name, layer "0" present: notification "Duplicated entry with name A found in LAYER", document read;
- no duplicate, no layer "0": document read, the default layer is created later.

Tested against master (3.7.11, a51b2108) and against v3.7.1, with the file above and with a real drawing that carries seven layers, one name twice and no "0".

### LINKED ISSUE #1242: `DxfReader.Read` throws `ArgumentException: An item with the same key has already been added. Key: 0` when a table contains a default entry twice
**Describe the bug**
`DxfReader.Read` throws `ArgumentException: An item with the same key has already been added. Key: 0` when a table contains a default entry twice (here layer `0` in the LAYER table), even with `Failsafe = true`.

The reader detects the duplicate and logs `Duplicated entry with name 0 found in LAYER`, then tries "last one wins" in `DxfTablesSectionReader.readEntries`:

```csharp
tableTemplate.CadObject.Remove(template.Name);
tableTemplate.CadObject.Add((T)template.CadObject);
```

`Table<T>.Remove` returns `null` without removing anything when the name is a default entry (`0`, `Continuous`, `Standard`, ...), so the following `Add` hits the dictionary key collision. Nothing in this path depends on the file version, it reproduces with AC1009 and AC1015 and with any duplicated default entry in any table. Affects 3.6.50 and 3.7.11 (current master).

Related to #1239 but not the same failure: #1239 duplicates a non-default entry (`A`) while the default entry is missing and fails inside `Remove` -> `assignToDefault` with a `KeyNotFoundException`. Here the duplicated entry *is* the default one and `Remove` never gets that far. A fix for one does not cover the other.

**To Reproduce**

```csharp
CadDocument doc = DxfReader.Read("LT_Modul_1.dxf", (s, e) => Console.WriteLine(e.Message));
```

```
[Warning] Duplicated entry with name 0 found in LAYER
System.ArgumentException: An item with the same key has already been added. Key: 0
   at ACadSharp.Tables.Collections.Table`1.add(String key, T item)
   at ACadSharp.Tables.Collections.Table`1.Add(T item)
   at ACadSharp.IO.DXF.DxfStreamReader.DxfTablesSectionReader.readEntries[T](CadTableTemplate`1 tableTemplate)
   at ACadSharp.IO.DXF.DxfStreamReader.DxfTablesSectionReader.readTable()
   at ACadSharp.IO.DXF.DxfStreamReader.DxfTablesSectionReader.Read()
   at ACadSharp.IO.DxfReader.readTables()
   at ACadSharp.IO.DxfReader.Read()
```

Minimal AC1015 file (the AC1009 form without handles and subclass markers behaves the same):

```
  0
SECTION
  2
HEADER
  9
$ACADVER
  1
AC1015
  0
ENDSEC
  0
SECTION
  2
TABLES
  0
TABLE
  2
LAYER
  5
2
100
AcDbSymbolTable
 70
2
  0
LAYER
  5
10
100
AcDbSymbolTableRecord
100
AcDbLayerTableRecord
  2
0
 70
0
 62
7
  6
Continuous
  0
LAYER
  5
11
100
AcDbSymbolTableRecord
100
AcDbLayerTableRecord
  2
0
 70
0
 62
7
  6
Continuous
  0
ENDTAB
  0
ENDSEC
  0
EOF
```

**Expected behavior**
With `Failsafe = true` the document loads, the first definition is kept and the already emitted warning is the only trace. With `Failsafe = false` the reader keeps throwing.

**Additional context**
Proposed fix, two small guards so that this issue and #1239 are covered together:

1. `DxfTablesSectionReader.readEntries`: skip the duplicate when `Remove` returns `null`.

   ```csharp
   if (tableTemplate.CadObject.Remove(template.Name) == null)
   {
   	//Default entries (layer "0", "Continuous", "Standard"...) cannot be removed, keep the first definition
   	continue;
   }
   ```

2. `Table<T>.assignToDefault`: return while the default entries do not exist yet, there are no references to reassign at that point (#1239).

   ```csharp
   if (!this.getDefaultEntries().All(this.entries.ContainsKey))
   {
   	return;
   }
   ```

The original file `LT_Modul_1.dxf` (AC1009, ASCII, R12 export of a laser-cutting tool) is attached.
[LT_Modul_1.zip](https://github.com/user-attachments/files/31820260/LT_Modul_1.zip)

---

**Release needed**

- [ ] Yes, using the Nuget package.
- [x] No, using the github repo.

dsn27:
# Description

Two guards in the DXF table reader so that a duplicated table entry no longer aborts the read when `Failsafe` is enabled.

- `DxfTablesSectionReader.readEntries` already detects a duplicated entry and replaces it ("last one wins"). `Table<T>.Remove` refuses default entries (`0`, `Continuous`, `Standard`, ...) and returns `null` without removing anything, so the following `Add` threw `ArgumentException`. The duplicate is now skipped and the first definition is kept. With `Failsafe = false` the reader still throws.
- `Table<T>.assignToDefault` resolved the default entry through the indexer while the table was still being read. When the default entry did not exist yet this threw `KeyNotFoundException`. There are no references to reassign at that point, so it now returns early.

Both cases are independent of the DXF version (reproduced with AC1009 and AC1015).

# Tasks done in this PR
* [x] `DxfTablesSectionReader.readEntries`: skip a duplicated default entry instead of throwing.
* [x] `Table<T>.assignToDefault`: return early while the default entries do not exist yet.
* [x] Samples `duplicated_default_entries_AC1009.dxf` and `duplicated_entry_no_default_AC1009.dxf`.
* [x] Regression tests `ReadDuplicatedDefaultLayerTest` and `ReadDuplicatedEntryWithoutDefaultTest`.

- Fixes #1242
- Fixes #1239

# Notes for reviewer
 - "Keep the first definition" was chosen for duplicated default entries because `Remove` cannot hand them out. For all other duplicates the existing "last one wins" behaviour is unchanged.
 - The `All(...)` check in `assignToDefault` is deliberately strict: it only skips the reassignment while the table is incomplete, i.e. during the read. On a fully built document the behaviour is unchanged.
 - `DxfReaderTests`, `DimensionTests`, `TableEntityTests` and `IOTests` pass locally on net9.0 (313 tests).

### DIFF
--- a/samples/duplicated_default_entries_AC1009.dxf
+++ b/samples/duplicated_default_entries_AC1009.dxf
@@ -1,0 +1,128 @@
+  0
+SECTION
+  2
+HEADER
+  9
+$ACADVER
+  1
+AC1009
+  0
+ENDSEC
+  0
+SECTION
+  2
+TABLES
+  0
+TABLE
+  2
+LAYER
+ 70
+2
+  0
+LAYER
+  2
+0
+ 70
+0
+ 62
+7
+  6
+CONTINUOUS
+  0
+LAYER
+  2
+0
+ 70
+0
+ 62
+7
+  6
+CONTINUOUS
+  0
+ENDTAB
+  0
+ENDSEC
+  0
+SECTION
+  2
+BLOCKS
+  0
+BLOCK
+  8
+0
+  2
+*D0
+ 70
+65
+ 10
+0.0
+ 20
+0.0
+  0
+LINE
+  5
+1
+  8
+Bemassung
+ 10
+0.0
+ 20
+5.0
+ 11
+100.0
+ 21
+5.0
+  0
+ENDBLK
+  8
+0
+  0
+ENDSEC
+  0
+SECTION
+  2
+ENTITIES
+  0
+LINE
+  5
+10
+  8
+Kanten
+ 10
+0.0
+ 20
+0.0
+ 11
+100.0
+ 21
+0.0
+  0
+DIMENSION
+  5
+11
+  8
+Konstruktion
+  2
+*D0
+ 10
+0.0
+ 20
+5.0
+ 11
+50.0
+ 21
+5.0
+ 70
+0
+ 13
+0.0
+ 23
+0.0
+ 14
+100.0
+ 24
+0.0
+  0
+ENDSEC
+  0
+EOF
--- a/samples/duplicated_entry_no_default_AC1009.dxf
+++ b/samples/duplicated_entry_no_default_AC1009.dxf
@@ -1,0 +1,66 @@
+  0
+SECTION
+  2
+HEADER
+  9
+$ACADVER
+  1
+AC1009
+  0
+ENDSEC
+  0
+SECTION
+  2
+TABLES
+  0
+TABLE
+  2
+LAYER
+ 70
+2
+  0
+LAYER
+  2
+A
+ 70
+0
+ 62
+7
+  6
+CONTINUOUS
+  0
+LAYER
+  2
+A
+ 70
+0
+ 62
+7
+  6
+CONTINUOUS
+  0
+ENDTAB
+  0
+ENDSEC
+  0
+SECTION
+  2
+ENTITIES
+  0
+LINE
+  5
+10
+  8
+A
+ 10
+0.0
+ 20
+0.0
+ 11
+100.0
+ 21
+0.0
+  0
+ENDSEC
+  0
+EOF
--- a/src/ACadSharp.Tests/IO/DXF/DxfReaderTests.cs
+++ b/src/ACadSharp.Tests/IO/DXF/DxfReaderTests.cs
@@ -2,6 +2,7 @@
 using ACadSharp.IO;
 using ACadSharp.Tests.TestModels;
 using System.Collections.Generic;
+using System.Linq;
 using Xunit;
 using Xunit.Abstractions;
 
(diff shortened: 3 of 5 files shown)


ilCosmico:
Hi @dsn27, I opened #1258 for the same two cases without noticing yours from two weeks ago, so I closed mine in favour of this one. I checked out this branch and it reads the file from #1239 on my side.

One detail worth keeping if the guard gets reviewed: the `All` in `assignToDefault` is doing real work. I had first written it as "return only when none of the default entries is there", and with that shape `LineTypes` still throws, since it has three default entries while `GetDefaultEntry` resolves `ByLayer` through the indexer. A table that holds `Continuous` but not `ByLayer` gets past the looser guard and throws inside `Remove`. Your version returns early there.

dsn27:
Hi @ilCosmico
Thanks for testing and for closing yours in favour of this one. 
Good point on the LineTypes case, that is exactly why the guard checks all default entries and not just any of them

PR #1245, opened 2026-09-04
Create layers referenced by DXF entities but missing from the LAYER table

### LINKED ISSUE #1244: DXF reader: create layers that are referenced by entities but missing from the LAYER table
**Is your feature request related to a problem? Please describe.**
Some exporters write DXF files whose entities reference layers (group code 8) that are not defined in the LAYER table. The file is invalid, but AutoCAD opens it and creates the missing layers on the fly (same as RECOVER). ACadSharp currently emits a warning in `CadEntityTemplate.build` and leaves the entity on layer `0`:

```
[Warning] ACadSharp.Tables.Layer table reference with handle:  | name: Kanten not found for ACadSharp.Entities.Line with handle 32
```

The layer name is gone from the document at that point. For us the layer carries the meaning of the geometry (cutting vs. engraving contours), so the drawing imports without errors but with every contour on layer `0`.

Layer references in DXF are names in every version, so this applies to all DXF versions (checked with AC1009 and AC1015).

Minimal AC1015 file: LAYER table with only `0`, one LINE on `Kanten`.

```
  0
SECTION
  2
HEADER
  9
$ACADVER
  1
AC1015
  0
ENDSEC
  0
SECTION
  2
TABLES
  0
TABLE
  2
LAYER
  5
2
100
AcDbSymbolTable
 70
1
  0
LAYER
  5
10
100
AcDbSymbolTableRecord
100
AcDbLayerTableRecord
  2
0
 70
0
 62
7
  6
Continuous
  0
ENDTAB
  0
ENDSEC
  0
SECTION
  2
ENTITIES
  0
LINE
  5
20
100
AcDbEntity
  8
Kanten
100
AcDbLine
 10
0.0
 20
0.0
 30
0.0
 11
100.0
 21
0.0
 31
0.0
  0
ENDSEC
  0
EOF
```

**Describe the solution you'd like**
When `getTableReference` finds neither the handle nor the name, but a name is present, create the layer with default properties, assign it to the entity and say so in the notification, e.g. in `CadEntityTemplate.build`:

```csharp
else if (!string.IsNullOrEmpty(this.LayerName) && builder.Layers != null)
{
	//The entity references a layer that is not defined in the LAYER table, create it instead of falling back to "0"
	builder.Notify($"Layer {this.LayerName} not found in the LAYER table, created for {this.CadObject.GetType().FullName} with handle {this.CadObject.Handle}", NotificationType.Warning);
	this.CadObject.Layer = builder.Layers.TryAdd(new Layer(this.LayerName));
}
```

Result on the minimal file: 2 layers, the line on `Kanten`. On the attached `LT_Modul_1.dxf` (AC1009, needs #ISSUE-1 to load): 5 layers, 17 lines on `Kanten`, 2 arcs and 2 lines on `Schneiden_Uhrzeiger`, 4 lines on `Konstruktion`, texts on `Bemassung`.

**Describe alternatives you've considered**
- A `CadReaderConfiguration` flag (e.g. `CreateMissingTableEntries`, default `false`) if the implicit creation should stay opt-in. The behaviour above would then only apply when the flag is set.
- Keeping the fallback to `0` but exposing the original name on the entity (e.g. via the notification args or an `UnresolvedLayerName`), so the consumer can recreate the layer. Works, but every consumer has to implement the same recovery.

[LT_Modul_1.zip](https://github.com/user-attachments/files/31821018/LT_Modul_1.zip)

dsn27:
# Description

When a DXF entity references a layer by name (group code 8) that is not defined in the LAYER table, `CadEntityTemplate.build` only emitted a warning and left the entity on layer `0`. The layer name was lost in the document.

With this change the reader creates the missing layer with default properties, adds it to the document's `Layers` table and assigns it to the entity, which is what AutoCAD does when it opens such a file. The warning stays and now says that the layer was created.

- Only applies when the reference could not be resolved by handle or by name and a name is present. 
- `Layers.TryAdd` is used, so several entities on the same missing layer share one `Layer` instance.

# Tasks done in this PR
* [x] `CadEntityTemplate.build`: create a layer that is referenced by name but missing from the LAYER table.
* [x] Sample `missing_layer_AC1009.dxf` (LAYER table with `0` only, two lines on `Kanten`, one circle on `Bohrungen`).
* [x] Regression test `ReadMissingLayerTest`.

- Closes #1244

# Notes for reviewer
 - This changes the default behaviour for invalid files: entities that used to land on `0` now keep their layer name. If you would rather keep the old behaviour as default, the block can be gated by a new `CadReaderConfiguration` flag. Happy to add that.
 - The created layer has default color, line type and flags. Nothing else about it is known from the file.

### DIFF
--- a/src/ACadSharp/IO/Templates/CadEntityTemplate.cs
+++ b/src/ACadSharp/IO/Templates/CadEntityTemplate.cs
@@ -71,6 +71,11 @@ internal class CadEntityTemplate : CadTemplate<Entity>
 		{
 			this.CadObject.Layer = layer;
 		}
+		else if (!string.IsNullOrEmpty(this.LayerName))
+		{
+			builder.Notify($"Layer {this.LayerName} not found in the LAYER table, created for {this.CadObject.GetType().FullName} with handle {this.CadObject.Handle}", NotificationType.Warning);
+			this.CadObject.Layer = builder.Layers.TryAdd(new Layer(this.LayerName));
+		}
 
 		switch (this.LtypeFlags)
 		{


src/ACadSharp/IO/Templates/CadEntityTemplate.cs
```diff
@@ -71,6 +71,12 @@ protected override void build(CadDocumentBuilder builder)
 		{
 			this.CadObject.Layer = layer;
 		}
+		else if (!string.IsNullOrEmpty(this.LayerName) && builder.Layers != null)
+		{
+			//The entity references a layer that is not defined in the LAYER table, create it instead of falling back to "0"
```
DomCR: This condition will never trigger for null, `builder.Layers != null` it can be removed.

The notification already describes what's going on, the comment can be removed.

src/ACadSharp.Tests/IO/DXF/DxfReaderTests.cs
```diff
@@ -180,4 +182,25 @@ public void ReadTables(FileModel test)
 		Assert.NotNull(doc.Views);
 		Assert.NotNull(doc.VPorts);
 	}
-}
\ No newline at end of file
+
+	[Fact]
+	public void ReadMissingLayerTest()
```
DomCR: The test file and case can be removed, I'll add it into the local testing, thanks.

src/ACadSharp/IO/Templates/CadEntityTemplate.cs
dsn27: Done, removed the null check and the comment.

src/ACadSharp.Tests/IO/DXF/DxfReaderTests.cs
dsn27: Done, test and sample removed. Thanks for taking it into the local tests.

PR #1251, opened 2026-09-08
fix(dxf): a DIMSTYLE record inherits the header value of an omitted code

simonedd:
The DIMSTYLE record is now seeded with the header variables so that only the codes actually present in the file overwrite them.


DomCR:
Hi @simonedd,

You are setting the dimensions using the override, not the default dimension, I'm not sure that this is the right behavior.

simonedd:
Hi @DomCR,

I think the property name is misleading here: `CadHeader.DimensionstyleOverrides` is not a DSTYLE
override, it is the storage of the header dimension variables themselves. I use them as default
values instead of the hard-coded `DimensionStyle` ones. 

My case: a DXF whose DIMSTYLE record omits 78 (DIMZIN) and 272 (DIMTDEC), while the header has `$DIMZIN 0` and `$DIMTDEC 4`. We read those two back as 8 and 2, the class hard-coded defaults; ODA reads them as 0 and 4.

